PR #11, opened 2015-03-19
ConceptInsights.semanticSearch ids as an array of strings instead of a JSON blob

### LINKED ISSUE #10: Concepts Insights - "ids" parameter of getConceptsMetadata()
I tried setting the "ids" parameter to:
1) just a string containing a single concept id
2) a string containing multiple concept ids separated by comma
3) an array of strings, each string for one concept id

but, each case, I got the following error:

{ error: 'invalid ids parameter', url: '' }

nfriedly:
This should fix #10 and also be a good pattern to follow for other APIs that expect JSON blobs in the HTTP request.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,16 @@
+# http://editorconfig.org/
+
+root = true
+
+[*]
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+indent_style = space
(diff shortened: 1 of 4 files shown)


nfriedly:
(I just re-did this: it now accepts the old style for backwards compatibility and also has tests that check it both ways.)

germanattanasio:
Thanks @nfriedly!

PR #27, opened 2015-06-24
Add authorization service wrapper

esbullington:
This is a PR for the new authorization service.  The Authorization service is structures like the other services, but only has a single function that sits on its prototype: getToken

This function takes a options object with a required 'url' property, since sending the URL of the intended resource for the Authorization service is required (tokens carry permissions that are constrained by path).

The other possibility is instead of passing in the URL, we could pass in the service that the token will be used with, which "knows" what the URL is.  Would this be a better approach?

I'm open to suggestions on improvements, and will check back by in a few hours (I need to spend a block of time now finishing up additional TTS demo app features).

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -92,7 +92,8 @@ var watson = {};
   'concept_insights',
   'tradeoff_analytics',
   'personality_insights',
-  'natural_language_classifier'
+  'natural_language_classifier',
+  'authorization'
 
 ].forEach(function(api) {
   watson[api] = createServiceAPI(api);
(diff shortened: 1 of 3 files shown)


esbullington:
Need to add test, one moment please.

germanattanasio:
@esbullington I already fixed the test case that is failing so don't worry about that

esbullington:
I can squash these if you'd like.

germanattanasio:
Let me merge this first

germanattanasio:
Can you add something in the readme to point out the new authorization service and how to use it ?

esbullington:
Sure, I'll send another PR in a few hours with that info added to the README (after I finish TTS stuff).

PR #33, opened 2015-07-02
Update STT and TTS examples and README

esbullington:
I'll also put in a PR for a fix to the TTS service here, there's still a '-beta' in the default link.

### DIFF
--- a/README.md
+++ b/README.md
@@ -479,7 +479,7 @@ var text_to_speech = watson.text_to_speech({
 
 var params = {
   text: 'Hello from IBM Watson',
-  voice: 'VoiceEnUsMichael', // Optional voice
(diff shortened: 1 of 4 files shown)


esbullington:
Actually, I'll just put that fix here, it will be easier for you.  One moment, please.

germanattanasio:
Eric we don't do that change until the service that was beta is removed.

esbullington:
OK, but then the updated examples do not work.  Is that OK?

germanattanasio:
For the examples we need to specify the url for now. then we will remove it

esbullington:
I'm submitting a new one with those changes, with your new commits merged.

PR #50, opened 2015-07-21
Add Solr search lifecycle and configuration client

gjs29:


germanattanasio:
Thanks for doing such a good job with the PR. I'm a little worried about not using the signature for requests that doesn't need parameters. 

You are assuming that the a method like `listClusters` will never have pagination or filtering. I know it's ugly but even for those methods we need to maintain the signature we use `method(params, callback)` It's not only more flexible but it also allow us to no prevent major releases because there is a signature change in the method.

What do you think?

germanattanasio:
@nfriedly what do you think ?

gjs29:
Yeah, using the same signature makes sense to me. I'll push up those changes in a bit.

germanattanasio:
Hi @gjs29, Thanks for the changes!

Issue #56, opened 2015-08-05
File called "string" created on startup


cfjedimaster:
I noticed my node app was creating a file called "string" with the contents: testtest

I narrowed it down to the init portion of this library:

var visual_recognition = watson.visual_recognition({
  username: 'stuff',
  password: 'more stuff',
  version: 'v1'
});

germanattanasio:
@cfjedimaster We don't create files in the wrapper/library. 
Are you using the sample app?

cfjedimaster:
Nope. I'm using this: https://github.com/cfjedimaster/IonicBluemixDemo

I basically commented out more and more of my app.js file and narrowed it down to that part.

nfriedly:
So, I just made all fs.write* functions log a stack trace, and it looks like it's coming from this: https://github.com/jprichardson/string.js/blob/master/lib/string.js#L726 which is in one of the watson-developer-cloud dependencies. I'm not exactly sure what they're trying to do there, but I don't think they're doing it, so I'm going to file a bug on that repo and link to this one.

cfjedimaster:
Muchas gracias.

On Wed, Aug 5, 2015 at 1:06 PM, Nathan Friedly <notifications@github.com>
wrote:

> So, I just made all fs.write* functions log a stack trace, and it looks
> like it's coming from this:
> https://github.com/jprichardson/string.js/blob/master/lib/string.js#L726
> which is in one of the watson-developer-cloud dependencies. I'm not exactly
> sure what they're trying to do there, but I don't think they're doing it,
> so I'm going to file a bug on that repo and link to this one.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/watson-developer-cloud/nodejs-wrapper/issues/56#issuecomment-128094205>
> .
>



-- 
===========================================================================
Raymond Camden, Developer Advocate for MobileFirst at IBM

Email : raymondcamden@gmail.com
Blog : www.raymondcamden.com
Twitter: raymondcamden

germanattanasio:
Thanks @nfriedly

jsstylos:
Nice debugging @nfriedly! Thanks for the report, @cfjedimaster.

nfriedly:
BTW, since @jsstylos asked about it, this is the code to make all fs.write* methods log a stack trace: https://gist.github.com/nfriedly/5813c0fcea944f05ccfe

nfriedly:
[...]

PR #57, opened 2015-08-05
update string version to get rid of #56

az7arul:


nfriedly:
Merged, thanks for the quick response @az7arul !

PR #58, opened 2015-08-10
fixed the documentation for the  authorization api

radleymith:
The callback for the authorization api getToken function takes the arguments (err, token), not just (token).  It's possible that the documentation should be changed further to reflect that so that the callback looks like

function (err, token) {
    if (err) {
        console.log('error:  ' + err);
    } else {
        //use token here
    }
}

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,7 +119,7 @@ var params = {
   url: 'https://stream.watsonplatform.net/text-to-speech/api'
 };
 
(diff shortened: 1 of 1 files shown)


germanattanasio:
Good catch @radleymith!

radleymith:
thanks, and thank you for answering my stackoverflow question, helped me out a lot.  I'm getting around to thumbs upping it etc.

PR #69, opened 2015-08-13
concept_insights/v2.js: adding return statements after callback(err).

m-m-f:
Concept insights v2.js was missing a return statement after all callbacks on error that were handled separately from the request helper.


germanattanasio:
Michele, add your ibm id to your github profile so that I know it's you :+1:

PR #75, opened 2015-08-26
Code review

germanattanasio:
I did a code review and found some unused variables.


nfriedly:
We should probably make `npm test` run jshint and then mocha. That would catch unused variables for us.

germanattanasio:
We should also ask people to read the CONTRIBUTING.md file :disappointed: 
I will look into the jshint + mocha or feel free to do it if is easy.

nfriedly:
I think it's easy, I'll try it out sometime soon here.

PR #78, opened 2015-08-31
Allow user config when creating a Solr cluster

gjs29:


services/retrieve_and_rank/v1.js
```diff
@@ -62,12 +62,33 @@ RetrieveAndRank.prototype.deleteClusters = function(params, callback) {
...
+ *   Optional params
+ *     - clusterName: name to use for identifying the cluster in responses
+ *     - clusterSize: size of the cluster to create
+ *     - clusterOptions: hash of other options used for creating the cluster
  *
  * @param callback The callback.
  */
 RetrieveAndRank.prototype.createCluster = function(params, callback) {
-  return sendRequest('POST', solrClustersPath(), this._options, callback);
+  if (params) {
```
germanattanasio: I would suggest to use something like:
```
params = params || {};
```
Right now you are not taking into account the scenario when users call `createCluster` without `params`

germanattanasio:
arguments should be `snake_case`

PR #90, opened 2015-09-14
updated README.md

timothyclifford:
updated README.md to use correct language_translation lib


nfriedly:
Good catch, thanks @timothyclifford

PR #92, opened 2015-09-14
fixed Tone Analyzer v2 synonyms API endpoint

herchu:
Fix to initial version of v2.js module: api endpoint changed to "synonyms"


nfriedly:
released in v0.10.1

PR #97, opened 2015-09-16

dolevdotan:


germanattanasio:
hi @dolevdotan, I can't accept your PR.

1. your changes are breaking the [error first callback](http://thenodeway.io/posts/understanding-error-first-callbacks/) pattern we use. 
1. your changes update the version in `package.json`

You need to put everything in the second parameter (`params`) and use `pick` and `omit`.
You can see how we do it in other services.

Something like:
```javascript
var omit = require('object.omit');
var pick = require('object.pick');

TradeoffAnalytics.prototype.dilemmas = function(params, callback) {
   var parameters = {
     options: {
       method: 'POST',
       url: '/v1/dilemmas',
      body: omit(params,['metadataHeader']),
      headers: {
    	'x-watson-metadata' : pick(params,['metadataHeader'])
      },
      json: true
     },
     requiredParams: ['columns', 'subject', 'options'],
     defaultOptions: this._options
   return requestFactory(parameters, callback);
 };
```

@jsstylos are you aware of `x-watson-metadata` and the new `/v1/events` call?

PR #98, opened 2015-09-16

dolevdotan:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "watson-developer-cloud",
-  "version": "0.10.3",
+  "version": "0.10.4",
```
germanattanasio: Versions are handle by a new release process. Please **don't modify it.**.
See [RELEASE.md](https://github.com/watson-developer-cloud/nodejs-wrapper/blob/master/RELEASE.md)

dolevdotan: Should I follow these steps? or are you going to do it?

germanattanasio: you should follow those steps.

nfriedly: @germanattanasio I would say that it should be your or me faccepting a PR and then ollowing the RELEASE.md steps. I don't think anyone else has permission to do the `npm publish` step.

germanattanasio:
ups sorry I was trying to say that @dolevdotan should follow: https://github.com/watson-developer-cloud/nodejs-wrapper/blob/master/CONTRIBUTING.md :( my bad.

PR #99, opened 2015-09-16

dolevdotan:
Per your suggestion, changed to conform to coding standards


germanattanasio:
Changes look good but the tests are failing. Probably because you don't check when params is null or undefined.

add this at the top of the method
```
params = params || {}
```

right after
```
 TradeoffAnalytics.prototype.dilemmas = function(params, callback) {
```

and please add a test case for events. See the test cases here: https://github.com/watson-developer-cloud/nodejs-wrapper/blob/master/test/test.tradeoff_analytics.v1.js

you can run the test cases with
``` npm test```

We can release things without test cases. specially knowing that >10k use this module.

dolevdotan:
closing in favor of pull request in dev branch

PR #100, opened 2015-09-17

dolevdotan:
OK, @germanattanasio, I've added tests, worked in dev, aligned with the holy coding guidelines, the whole nine yards...
Please accept the pull request so we can get this over with...


germanattanasio:
Thanks @dolevdotan! the code looks pretty solid. Sorry for being too picky. 

I will update the nodejs later today along with the demo

PR #105, opened 2015-09-22
Add support for rankers in retrieve and rank

germanattanasio:
@kshasan will be testing the changes and updating the documentation. We should not merge this until he is done.


germanattanasio:
@kshasan seems like there are some broken tests

PR #116, opened 2015-10-01
Update default url for R&R service

gjs29:


germanattanasio:
I think this is fixed in dev and will be merged in the next release

germanattanasio:
See https://github.com/watson-developer-cloud/nodejs-wrapper/blob/dev/services/retrieve_and_rank/v1.js

PR #122, opened 2015-10-11
update retrieve and rank

germanattanasio:
### Update the retrieve and rank service implementation

@gjs29 Can you verify if `deleteCollection`, `createCollection` and `listCollections` are actually `GET` calls and not `POST` as described in:

* [swagger](http://www.ibm.com/smarterplanet/us/en/ibmwatson/developercloud/apis/#!/retrieve-and-rank/adminAction)
screen shot 2015-10-11 at 12 40 39 pm

* [API Reference](http://www.ibm.com/smarterplanet/us/en/ibmwatson/developercloud/retrieve-and-rank/api/v1/#delete_solr_collection).
screen shot 2015-10-11 at 12 40 28 pm

I'm looking at the [Solr documentation](https://cwiki.apache.org/confluence/display/solr/Collections+API) and looks like those are `GET` requests but I want to make sure.

### DIFF
--- a/services/retrieve_and_rank/v1.js
+++ b/services/retrieve_and_rank/v1.js
@@ -444,7 +444,7 @@ RetrieveAndRank.prototype.createCollection = function(params, callback) {
   var parameters = {
     options: {
       url: '/v1/solr_clusters/{cluster_id}/solr/admin/collections',
-      method: 'GET',
+      method: 'POST',
       qs: queryParams,
       path: pick(params,['cluster_id']),
       json: true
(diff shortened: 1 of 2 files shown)


gjs29:
Solr (and our service) will accept both POSTs and GETs for collection actions.

germanattanasio:
should we use POST or GET in the sdk ?

nfriedly:
Use GET when you're only fetching information (not making changes) and you can be fairly certain that the URL won't exceed ~2000 characters

gjs29:
That sounds like GET for listCollections and POST for create and delete.

nfriedly:
Yep

germanattanasio:
I made delete and create to be POST

PR #130, opened 2015-10-27
Merge dev into v1.0

germanattanasio:
This will merge the integration tests that were added to `dev` into our major release


germanattanasio:
@nfriedly I'm merging dev so that we get the integration tests and some examples from the document conversion team

PR #148, opened 2015-11-16
corrected usage of visual_recognition service

jsloyer:


jsstylos:
Why are we switching from "v1-beta" back to "v1"?

germanattanasio:
OMG. I just realized that this is going back to `v1` :( my bad

PR #153, opened 2015-11-18
Adds configuration example for the html-to-html conversion step

tdumitriu:
By adding this configuration example we want to show developers
the way they can set attributes to tune the document conversion.


nfriedly:
Thanks!

PR #171, opened 2015-12-04
Updates travis commands for better PRs

mfulgo:
This should fix the failed Travis builds when checking Pull Requests.

It should end up skipping the integration tests, but the unit tests
should still run.


nfriedly:
Thanks, nice work!

PR #174, opened 2015-12-09
Adds var fs for personality-insights

rastringer:
Hi German, 

 Thanks for these fantastic Watson services. This is a tiny PR -- I just noticed that the personality_insights.vs.js file needs a var fs = require('fs')  in order for the  }).pipe(fs.createWriteStream('./output.csv'));  to run at the end. All tests are passing. 

Hope you're well, Robin


germanattanasio:
thanks!

PR #188, opened 2015-12-21
support for keywords, keywords_threshold, and word_alternatives_thres…

monte-hayward:
The v1 service wrapper would `pick` params allowed to be sent to the service. This pr updates the "allowed" list with the new parameters announced in [Updates for December 2015](http://www.ibm.com/smarterplanet/us/en/ibmwatson/developercloud/doc/speech-to-text/upgrade.shtml).

adds support for keywords, keywords_threshold, and word_alternatives_threshold

### DIFF
--- a/services/speech_to_text/v1.js
+++ b/services/speech_to_text/v1.js
@@ -31,6 +31,9 @@ var util           = require('util');
 var WebSocketClient = require('websocket').client;
 var pkg         = require('../../package.json');
 
+const PARAMS_ALLOWED = ['continuous', 'max_alternatives', 'timestamps', 'word_confidence', 'inactivity_timeout',
+  'model', 'content-type', 'interim_results', 'keywords', 'keywords_threshold', 'word_alternatives_threshold' ];
+
 function formatChunk(chunk) {
   // Convert the string into an array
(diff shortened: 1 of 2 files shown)


monte-hayward:
recognizeStream was untested before, and now has some tests. 
Imo, it has more coverage than before the commit.
Maybe the Coveralls fail is due to [coveralls@2.11.5 breaks build #176](https://github.com/watson-developer-cloud/node-sdk/pull/176)?
Thanks

services/speech_to_text/v1.js
```diff
@@ -394,8 +397,7 @@ function RecognizeStream(options){
     'content-type': 'audio/wav', // todo: try to determine content-type from the file extension if available
     'continuous': false,
     'interim_results': true
-  }, pick(options, ['continuous', 'max_alternatives', 'timestamps',
-    'word_confidence', 'inactivity_timeout', 'content-type', 'interim_results']));
+  }, pick(options, [PARAMS_ALLOWED]));
```
germanattanasio: I think this line should be 
```
pick(options, PARAMS_ALLOWED));
```

PR #199, opened 2016-01-08
[speech-to-text] Add recognize options

monte-hayward:
Also threw in some test resources.
This is the file that threw the error:
     test/resources/malformed.flac
I believe I misreported the bug; I'm not sure what result would occur with silence or noise. Included those in case you want to write more tests.

Thanks, @nfriedly ! 
I should have noted that I was working on it.


services/speech_to_text/v1.js
```diff
@@ -391,13 +391,7 @@ function RecognizeStream(options){
 
   var queryParams = extend({model: 'en-US_BroadbandModel'}, pick(options, ['model', 'X-Watson-Learning-Opt-Out', 'watson-token']));
 
-  var openingMessage = extend({
-    // todo: confirm the mixed underscores/hyphens and/or get it fixed
-    action: 'start',
-    'content-type': 'audio/wav', // todo: try to determine content-type from the file extension if available
-    'continuous': false,
-    'interim_results': true
-  }, pick(options, [PARAMS_ALLOWED]));
```
nfriedly: Isn't some of this needed?

nfriedly:
Thanks again, the fix was released in v1.1.3 just now.

PR #214, opened 2016-02-22
Fix query param to be tones not tone

jthomas:


germanattanasio:
I bet this was a headache for you @jthomas

jthomas:
My own fault :) I'm an idiot. I blame coding on an :airplane:

Issue #230, opened 2016-04-14
[retrieve-and-rank] uploadConfig API call does not work in Bluemix


DanCunnington:
The uploadConfig API call for Retrieve and Rank can't be used in Bluemix as it requires a .zip file path instead of a .zip file.

http://www.ibm.com/smarterplanet/us/en/ibmwatson/developercloud/retrieve-and-rank/api/v1/?node#upload_config

nfriedly:
I believe this is also fixed, please re-open if I'm wrong.

DanCunnington:
Yes this is fine, I used the 'temp' npm module as a workaround for Bluemix environments

PR #231, opened 2016-04-14
uploadConfig to use Buffer parameter passed in

DanCunnington:
Fix for issue #230


DanCunnington:
Closing due to build failures and also @germanattanasio allowing a stream to be passed in similar to speech to text

PR #232, opened 2016-04-15
Config path

germanattanasio:
Add support for #230


germanattanasio:
I just tested and seems like the changes are working. I'm going to merge this so that @DanCunnington can work on the node-red nodes

PR #236, opened 2016-04-26
Move "url" attributes to fix endpoint selection

g-may:
### Summary

getFormat in helper.js should not immediately associate the "url" parameter with an Alchemy URL API call. Alchemy HTML and Text APIs can also pass "url" parameters.

### Other Information

[helper.js](https://github.com/watson-developer-cloud/node-sdk/blob/master/lib/helper.js)
``` javascript 
  /**
   * Returns the first match from formats that is key the params map
   * otherwise null
   * @param  {Object}  params   The parameters
   * @param  {Array}  requires The keys we want to check
    */
  getFormat: function(params, formats) {
    if (!formats || !params)
      return null;

    for(var i = 0; i < formats.length; i++) {
      if (formats[i] in params)
        return formats[i];
    }
    return null;
  }
```
[alchemy_language/v1.js](https://github.com/watson-developer-cloud/node-sdk/blob/master/services/alchemy_language/v1.js)
``` javascript
var endpoints      = require('../../lib/alchemy_endpoints.json');
...
function createRequest(method) {
  return function(_params, callback ) {
    var params = _params || {};
    var accepted_formats = Object.keys(endpoints[method]);
    var format = helper.getFormat(params, accepted_formats);

```

### DIFF
--- a/lib/alchemy_endpoints.json
+++ b/lib/alchemy_endpoints.json
@@ -1,116 +1,116 @@
 {
   "entities" : {
-    "url"  : "/url/URLGetRankedNamedEntities",
     "text" : "/text/TextGetRankedNamedEntities",
-    "html" : "/html/HTMLGetRankedNamedEntities"
+    "html" : "/html/HTMLGetRankedNamedEntities",
+    "url"  : "/url/URLGetRankedNamedEntities"
   },
 
   "keywords": {
-    "url"  : "/url/URLGetRankedKeywords",
     "text" : "/text/TextGetRankedKeywords",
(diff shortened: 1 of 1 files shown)


nfriedly:
I think this is OK to merge since it fixes a real bug, but it feels like it might get broken again fairly easily. Do you think you could also add in some unit tests and comments to validate the correct behavior and explain what's going on there?

Or, alternatively, could we make a change (perhaps rename `url` to `endpoint_url`) to make this a little safer?

germanattanasio:
can you do this @g-may ?

PR #237, opened 2016-04-27
Add tests for Alchemy endpoint selection

g-may:
### Summary

Tests for [pull 236](https://github.com/watson-developer-cloud/node-sdk/pull/236)

### Other Information

Thanks for contributing to the Watson Developer Cloud!


nfriedly:
Cool, thanks @g-may!

BTW, you can add more commits to the first branch and they will automatically show up on the original pull request (that way it would show the tests as passing). But this is fine for today.

Issue #240, opened 2016-05-05
[language-translation] Example on the readme.md page for language translation gives an error 


prashantgupta24:
The error is:  
      
``` 
{
  error_code: 404,
  error_message: 'cannot find service matching the request data'
}
```

It seems to be having an extra parameter, which is not present in the example on language translation in the docs.

Suggested solution is to remove the extra parameter present in the example on the readme.md page

nfriedly:
Fixed in cafcec4

PR #241, opened 2016-05-05
Updated language translation example

prashantgupta24:
### Summary

Example on the readme.md page for language translation gives an error. Issue #240
The language translation function call had an extra parameter, which was causing problems in using the functionality.

### Other Information
In the example on the main page, the *model-id* field is causing problems:
```
language_translation.translate({
  text: 'A sentence must have a verb', source : 'en', target: 'es', model_id: '<model-id>' },
  function (err, translation) {
    if (err)
      console.log('error:', err);
    else
      console.log(JSON.stringify(translation, null, 2));
});
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -316,7 +316,7 @@ var language_translation = watson.language_translation({
 });
 
 language_translation.translate({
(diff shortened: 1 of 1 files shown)


germanattanasio:
thanks!

PR #248, opened 2016-05-16
update version date

jzhang300:


Kikobeats:
@jzhang300 a little question related with `version_date` parameter, why it is necessary? I'm working with visual recognition, I need to pass this value or just it is for testing?

Kikobeats:
In my case

```
Error: Argument error: version_date was not specified, use 2015-12-02
```

But what this values means?

germanattanasio:
The parameter is required by the API. See https://github.com/watson-developer-cloud/api-guidelines#versioning

You can set this parameter in the constructor.

```js
var tone_analyzer = watson.tone_analyzer({
  username: 'USERNAME FOR THE SERVICE HERE',
  password: 'PASSWORD FOR THE SERVICE HERE',
  version: 'v3',
  version_date: '2016-05-19'
});
```

PR #250, opened 2016-05-17
Update README: Change tone analyzer `v3` to `v3-beta`

kauffecup:
### Summary

There is no `tone_analyzer v3` version. This change updates the README to point to `v3-beta` instead.

Avoids error:

```sh
Error: Service tone_analyzer v3 not found. Did you mean v3-beta?
```


germanattanasio:
@kauffecup we are working on tone v3. It will be release soon

kauffecup:
I understand that, but in the meantime why not keep the README up to date so that developers using the service don't hit errors?

PR #251, opened 2016-05-18
Added visual recognition v3

germanattanasio:


nfriedly:
Things to do / check:

- [ ] Fix jsdoc
- [ ] write new tests
- [ ] ensure it can use `VCAP_SERVICES` - it's now `api_key` instead of `username` and `password`

PR #267, opened 2016-06-13
[visual-recognition] Return error if invalid api_key, fixes #264

### LINKED ISSUE #264: [visual-recognition] Invalid API key message returned in results argument, not error argument
When the visual recognition API is called with an invalid API key, an error is returned:

    { status: 'ERROR', statusInfo: 'invalid-api-key' }

However, this should be in the error argument, not the results argument for the callback function:

```js
var path = require('path');
var fs = require('fs');
var watson = require('watson-developer-cloud');

var visualRecognition = watson.visual_recognition({
  api_key: 'bogus',
  version: 'v3',
  version_date: '2016-05-20'
});

visualRecognition.classify({
  images_file: fs.createReadStream(path.join(__dirname, '../../', 'public/images/bundles/dogs/test/1.jpg'))
}, function(err, res) {
  // expected: err should contain the error message, res should be null
  // actual: err is null and res contains the error message
  console.log(err, res);
});
```

germanattanasio:


nfriedly:
@germanattanasio This appears to break the visual recognition v2-beta integration tests.. or is that service just sunset now?

germanattanasio:
I didn't change anything in the v2 :(

nfriedly:
Yea... let me see what happens locally

nfriedly:
Yea, it fails for me locally on an up-to-date master. I'm going to merge this, but we should figure out what's up with v2-beta.

PR #270, opened 2016-06-16
allow having http headers in Tone Analyzer wrapper options

herchu:
### Summary

Allows instantiating a Tone Analyzer service with customer http headers.

### Other Information

Current code overrides all headers. With this code (only change, uses `extend(true, ...))` one can instantiate the service for example:
```
var toneAnalyzer = watson.tone_analyzer({
  url: 'https://gateway.watsonplatform.net/tone-analyzer/api/',
...
  headers: { 'Accept-Language': 'es' }
})
```
and have these headers honored. This is the only way to get internationalized values through the node sdk.

### DIFF
--- a/services/tone_analyzer/v3.js
+++ b/services/tone_analyzer/v3.js
@@ -74,11 +74,11 @@ ToneAnalyzer.prototype.tone = function(params, callback) {
       body: params.text,
       qs: pick(params, ['tones', 'sentences'])
     },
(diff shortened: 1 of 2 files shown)


nfriedly:
Commit looks good; do you think you can add a unit test with nock to match it?

herchu:
@nfriedly Test added - can you review? Thanks!

nfriedly:
Perfect, thanks!

PR #277, opened 2016-07-05
[document-conversion] Adds index document API

samir-patel:
### Summary

This submission includes new API call for the Document Conversion Service that will allow you to access the index_document endpoint:

DocumentConversion.prototype.index

This is a synchronous call that will allow you to convert and index a document into a SOLR collection (via the Retrieve and Rank service) in one operation.  A "dry run" of the indexing can be submitted to preview the indexed results prior to actual indexing.

### DIFF
--- a/examples/document_conversion.v1.js
+++ b/examples/document_conversion.v1.js
@@ -22,6 +22,137 @@ document_conversion.convert({
     }
   }
 }, function (err, response) {
+  console.log("----------\n");
+  console.log("convert a single document\n");
+  console.log("----------\n");
+  if (err) {
+    console.error(err);
+  } else {
+    console.log(JSON.stringify(response, null, 2));
+  }
+});
+
+// dry run of indexing a single document
+document_conversion.index({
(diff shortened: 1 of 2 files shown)


germanattanasio:
@nfriedly do we get `lint` to check that the `jsdoc` is valid ?

germanattanasio:
This looks OK for me

nfriedly:
@germanattanasio we have the [ESLlint rule disabled](https://github.com/watson-developer-cloud/node-sdk/blob/master/.eslintrc.js#L206) but we do run the [doc build on travis](https://github.com/watson-developer-cloud/node-sdk/blob/master/.travis.yml#L9) as part of the tests and it looks like [this one passed](https://travis-ci.org/watson-developer-cloud/node-sdk/jobs/142507867#L836)

nfriedly:
Hey @samir-patel, sorry it took me a couple of days to get to this; do you think you can add a basic "happy case" unit test?

PR #282, opened 2016-07-13
Change current version date and set conversion target to lower case

SirSpidey:
Version date default is 2015-12-15.

Examples fail with current lower case conversion_targets

### Summary
* updates from 2015-012-01 to 2015-12-15
* changes conversion target to lowercase. See
http://www.ibm.com/watson/developercloud/doc/document-conversion/updates
.shtml#previous and search for **14 March 2016**

### Other Information
I suspect that you'll want to change the code to support both uppercase and lowercase conversion targets. Tell me where to put that and I'll add that.

### DIFF
--- a/services/document_conversion/v1.js
+++ b/services/document_conversion/v1.js
@@ -28,13 +28,13 @@ var omit           = require('object.omit');
  */
 function DocumentConversion(options) {
   // Warn if not specifying version date
-  var version_date = '2015-12-01';
+  var version_date = '2015-12-15';
   if(options && options.version_date) {
     version_date = options.version_date;
   } else {
     // eslint-disable-next-line no-console
     console.warn('[DocumentConversion] WARNING: No version_date specified. Using a (possibly old) default. ' +
-                  'e.g. watson.document_conversion({ version_date: "2015-12-01" })');
+                  'e.g. watson.document_conversion({ version_date: "2015-12-15" })');
   }
 
   // Default URL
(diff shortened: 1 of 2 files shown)


nfriedly:
To support both, right in between these two lines:

```js
   params = params || {};
   if (!params.conversion_target || !DocumentConversion.prototype.conversion_target[params.conversion_target]) {
   if (!params.conversion_target || !DocumentConversion.prototype.conversion_target[params.conversion_target]) {
```
you could add something like

```js
if (typeof params.conversion_target === 'string') {
    params.conversion_target = params.conversion_target.toLowerCase();
}
```
(and then an extra unit test to make sure it works either way.)

We should probably also update [these guys](https://github.com/watson-developer-cloud/node-sdk/blob/master/services/document_conversion/v1.js#L51) (just the values, the keys are fine).

```
DocumentConversion.prototype.conversion_target = {
  ANSWER_UNITS: 'ANSWER_UNITS',
  NORMALIZED_HTML: 'NORMALIZED_HTML',
  NORMALIZED_TEXT: 'NORMALIZED_TEXT'
};
```

Finally, why change the default version date?

SirSpidey:
@nfriedly: I changed the version date to match what was announced in the release notes and what we have in the [docs](http://www.ibm.com/watson/developercloud/doc/document-conversion/updates.shtml#changelog):
>The current version is 2015-12-15, and the documentation reflects the current version.

nfriedly:
Also, it looks like a bunch of the existing doc conversion tests are failing now. Any ideas?  https://travis-ci.org/watson-developer-cloud/node-sdk/jobs/144778583#L952

SirSpidey:
@nfriedly: Other than my bad coding or that our tests are not going against the current GA service, I don't know.  You can see from the API explorer or a curl command that the lowercase conversion_target is ok: 

```
curl -X POST -u "{username}":"{password}" -F config="{\"conversion_target\":\"answer_units\"}" -F "file=@sample-docx.docx" "https://gateway.watsonplatform.net/document-conversion/api/v1/convert_document?version=2015-12-15"
```

nfriedly:
Ok, I'll dig in in a little bit and see if I can figure out why.

nfriedly:
I figured it out. The `convert()` method was checking `params.conversion_target` against the `DocumentConversion.prototype.conversion_target` keys (uppercase) rather than the values (lower case). Previously they were both upper case, so it didn't matter.

PR #287, opened 2016-07-25
Add examples for AlchemyLanguage

g-may:
Adding examples for the rest of the functions


kognate:
The test failures seem to be mostly eslint issues.  Once those are fixed,  I think this will be OK.

Issue #283, opened 2016-07-16
SDK does not work when run through webpack


bradennapier:
This service is unbelievably hard to get started with.  Your links aren't working, when you add a credit card you can't get back to console without searching on google or knowing the link, services are difficult and confusing to navigate...

Now to the actual problem... I created a service and got a JSON for credentials taht does NOT include any kind of version.  There is zero indication of "version" anywhere on any documentation for your TTS service anywhere.  You ahve NOTHING in any docs indicating how to get this version information... so I use the example since it is TTS and use v1...

I have checked the login and password on my bluemix console 10,000 times - and yet I get an error the second I call 

```
watson.authorization({
     username: '<my username>',
    password: '<my password>',
    version: 'v1'
})
```

`
Service authorization v1 not found.`

nfriedly:
Hi @bradennapier,

I'm sorry to hear you're having so much trouble.

> Your links aren't working,

Can you point me to which specific links aren't working?

> ` Service authorization v1 not found.`

Are you you trying to use the code via browserify/webpack/etc? The current version of the Node.js SDK only works in Node.js, however client-side support has been frequently requested, and I've been [working on it](https://github.com/watson-developer-cloud/node-sdk/tree/browser) for the past week or so. (It's not ready yet, though.)

For Text to Speech in Web Browsers, please use the [watson-speech](https://www.npmjs.com/package/watson-speech) package for the moment - it's browserify friendly and has [lots of examples](https://github.com/watson-developer-cloud/speech-javascript-sdk/tree/master/examples/static/), including a [node.js server](https://github.com/watson-developer-cloud/speech-javascript-sdk/blob/master/examples/server.js) to run the Authorization service and generate access tokens. (The authorization service must be used on the server rather than on the client, otherwise it'll leak your credentials, which defeats it's main purpose.)

Again, I'm sorry for all of the trouble. A lot of this stuff is really new and I know there's lots of room for improvement. Myself and many others are working very hard to get it there. Please give this a shot, and if it doesn't resolve your issues, let me know what your goal is and where you're getting stuck.

bradennapier:
Well I am using that as well but the node sdk's authorization does not work server side.  I am running it via Node 4.3 on Amazon Lambda AWS and get that error every time.

Serivde authorization v1 not found occurs the second I call

``` watson.authorization({ ...authData }) ```

i changed to using the request package and a simple HTTP get and it works... not a good experience

**Comment by bradennapier** [comment]
```
[...]

Issue #257, opened 2016-05-31
[speech-to-text] Change createRecognizeStream() defaults to match service defaults


jeffpk62:
The createRecognizeStream() method currently hard-codes parameter values.  Lines 72 through 81 of https://github.com/watson-developer-cloud/node-sdk/blob/master/services/speech_to_text/recognize_stream.js are as follows:

```
 var openingMessage = extend({
    action: 'start',
    'content-type': 'audio/wav',
    continuous: true,
    interim_results: true,
    word_confidence: true,
    timestamps: true,
    max_alternatives: 3,
    inactivity_timeout: 600
  }, pick(options, OPENING_MESSAGE_PARAMS_ALLOWED));
```

These settings override whatever the user passes in via the parameters of the method.  We need to eliminate the hard-coding and honor the params value passed by the caller.

jeffpk62:
Nathan, you are correct about overrides:  I can override the hard-coded settings by setting the parameters to their default values.  In the SDK, we shouldn't turn on anything unless requested by the user to conform to the expected default behavior for the service.  With no param values, the user should expect the same default behavior from all interfaces.

I did more testing with the audio formats, and there is something odd going on.  With the recognize() method, if I set the content_type to audio/wav and send a .flac file, I get the following error:

```
{ code_description: 'Bad Request',
  code: 400,
  error: 'unable to transcode data stream audio/wav -> audio/x-float-array ' }
```

But doing the same thing with createRecognizeStream() produces successful results, as if I'd not set a content_type that differs from the actual type of the input file.

nfriedly:
I think this change may have to wait until v3 is published, but fair enough. I don't have an exact date, but I'm thinking sometime in the next few months. (It'll have a few breaking changes, the largest of which will be geared towards working in web browsers.)

The browser Speech JS SDK is more reserved here, it enables `continuous` by default because I've seen a lot of confusion around that param (so much so that believe it may be removed entirely, or at least have the default changed in a future version of the service), and it enables `interim_results` because otherwise the service seems to be buffering all results to the end otherwise (even over websockets - which basically defeats the purpose of websockets).

jeffpk62:
[...]

PR #288, opened 2016-07-27
Browser support

### LINKED ISSUE #30: The library should support authenticating with token
Developers should be able to create an instance of the service when they only have a token, not just a username and password.

### LINKED ISSUE #204: Consider a package that works from browser or enabling CORS
Currently the service doesn't work with cross-domain requests and there is no package that works in the browser.  This limits the users who can actually test the service.

### LINKED ISSUE #225: Restructuring plan for browser support
Several Watson services currently support CORS, and others will soon (hopefully). Users will [try to use the Node.js SDK in a web browser](http://stackoverflow.com/questions/35952647/watson-visual-recognition-run-error/36020024) (via browserify, etc.) but it fails with "file not found" sorts of errors when trying to create a service instance due to the way that we programmatically require modules on demand.

A simple fix might be to just `require()` everything at the top level, and this should probably be part of the strategy so that we can generate a single bundle.js for browsers that includes all supported Watson services (and helpful error messages when a user attempts to use services that do not support CORS in a browser - this is easy to automate with browserify).

However, I think this is a good opportunity to take things a step further:

* Allow users to `require('watson-developer-cloud/some-service/v1')` so that they can create their own bundles with only the code they're using.

* The above step requires a slight re-organizing of the code (moving service folders out of services/ to the top level), so it may be a good time to re-organize the readme, examples, and tests, and group things by service instead of type. 

  The readme in particular is getting a bit hairy. Having a separate readme for each service could help. (Alternatively, we could just move things into the comments, and have the readme link to the JSDoc... I'm not as sure about this point, but wanted to bring it up for discussion.)

* Switch to ES6 modules to allow for tree-shaking (even smaller file size) and nicer syntax (we can also include pre-compiled copies of everything in the npm releases - deployment is automated, so there aren't any extra manual steps to remember.)

* Combine the Node & Browser SDKs into a single "JavaScript SDK" that runs everywhere

@germanattanasio, @jsstylos, & @jzhang300: what do you guys think?

nfriedly:
This still needs a little cleanup before shipping, including:

* Update readme and examples to new constructor style
* Some automated in-browser tests
* More details on which services support CORS and which don't
* Some time spent berating the services that don't support CORS

But, I'm creating the PR now in hopes of getting some review from @germanattanasio, @jzhang300, @kognate, and anyone else interested in commenting.

## Summary of changes

* I re-organized the code so that all services can be `require()`'d and instantiated individually - this allows it to be compatible with browserify and also enables smaller bundles that don't include the entire library
  * In the process of this, I created a base class that had the general logic from the previous `lib/index.js`, but allowed service-specific hacks to be moved into their own files.

* I changed the order or preference for credentials and added a new option for simple env properties.

* I gave the Authorization service a little love so that you can now drop in the credentials object from VCAP_SERVICES to the constructor, and then call `getToken()` without specifying the URL again.

* I also put the default URL for each service as a static property on it's constructor to make it easy to pass to the Auth service in non-bluemix environments.

## Things I didn't do in this PR, but might like to slip into a v2 release:

* Promises. Ideally, I'd like everything to return Promises instead of streams unless a stream is specifically requested, and possibly not even then unless it actually makes sense (e.g. speech services). This provides potential to build a much lighter-weight client side (particularly if we don't let it support streams at all.)

* Merging in the Speech JS SDK & renaming this to the JavaScript SDK
 
* Removing `request`-specific features. We lean pretty heavily on request, but being able to switch it out with something like [axioz](https://www.npmjs.com/package/axios) or possibly just using [fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) client-side and a [pollyfill](https://www.npmjs.com/package/node-fetch) in node would be handy.

* Reducing the size of the bundle - I think it weighs in at like 3.7mb right now (uncompressed - it's like 370kb when minified and gzipped, but still...) Dropping request in favor of fetch would help here.

* Automatically generating a complete client-side bundle for GithubReleases. Pretty easy to do, but I'd like to get the file-size down a bit first.

* Dropping support for node `v0.12`. It's near EOL, and would allow us to use a bit of es6 sugar in the library. Not a bit deal to me though.

* Breaking changes to make default settings match service defaults. https://github.com/watson-developer-cloud/node-sdk/issues/257 is the only one I know of off the top of my head, but there may be a few others.

Some of this can slip to `v3` - I'd be interested in feedback on what you think makes sense for `v2` vs waiting on.

## Tickets

This PR resolves https://github.ibm.com/Watson/developer-experience/issues/234 and https://github.ibm.com/Watson/developer-experience/issues/311 on GHE.

Also, on this repo, it
Fixes #30 
Fixes #225 
Fixes #204 
Relates to #283


germanattanasio:
Changes look OK for me. Great job btw.

What if we remove the [solr-client](https://www.npmjs.com/package/solr-client) dependency. We can ask users to install it if they want to use Retrieve and Rank but having it as a dependency for all our users  doesn't make sense.

nfriedly:
I don't think I agree about solr-client. If it's necessary to use that part of the library, then it should be included as a dependency automatically.

But, if a user does, for example `require('watson-developer-cloud/conversation/v1')`, then they will have solr-client installed in their `node_modules/`, but it won't be loaded at runtime (or included in a browserify bundle).

As for excluding non-CORS services from browserify.. that's not a bad idea, but I think we can punt it for now. (And hope that they all support CORS in the near future...)

germanattanasio:
makes sense.

PR #292, opened 2016-08-01
Fixes issue with retraining function

kognate:
The retrain function was concatinating keys badly if only
one field was submitted. There was another concat problem with the
classifier_id


germanattanasio:
ups

PR #296, opened 2016-08-09
integration of tone and conversation example code

aprilwebster:
### Summary
Sample code for the integration of tone and conversation as per discussion on August 1.

### Other Information
N/A.


.project
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
germanattanasio: please remove this file

examples/conversation_addons/tone_detection.js
```diff
@@ -0,0 +1,281 @@
...
+var watson = require('watson-developer-cloud');
+var Promise = require('bluebird');
+
+/**
+ * Thresholds for identifying meaningful tones returned by the Watson Tone Analyzer.  Current values are
+ * based on the recommendations made by the Watson Tone Analyzer at
+ * https://www.ibm.com/watson/developercloud/doc/tone-analyzer/understanding-tone.shtml
+ * These thresholds can be adjusted to client/domain requirements.
+ */
+var PRIMARY_EMOTION_SCORE_THRESHOLD = parseFloat("0.5");
```
germanattanasio: use `'` instead of `"`

examples/conversation_addons/tone_detection.js
```diff
@@ -0,0 +1,281 @@
...
+ * @param conversationPayload json object returned by the Watson Conversation Service
+ * @param toneAnalyzerPayload json object returned by the Watson Tone Analyzer Service
+ * @returns conversationPayload where the user object has been updated with tone information from the toneAnalyzerPayload 
+ */
+function updateUserTone (conversationPayload, toneAnalyzerPayload) {
+  var emotionTone = null;
+  var languageTone = null;
+  var socialTone = null;
+
+  if(typeof conversationPayload.context === 'undefined')
```
germanattanasio: Open brackets in the same line the `if` is declared. See our [CONTRIBUTING](https://github.com/watson-developer-cloud/node-sdk/blob/master/.github/CONTRIBUTING.md) guidelines

germanattanasio:
@aprilwebster Thanks for contributing to the node-sdk, few things:

- [ ] Add your IBM email to your GitHub profile so that it match the commits with your GitHub account.
- [ ] ask Vibha Sinha to do the same and sign the CLA 
- [ ] Update the code to follow the [CONTRIBUTING](https://github.com/watson-developer-cloud/node-sdk/blob/master/.github/CONTRIBUTING.md) guidelines.

germanattanasio:
@nfriedly this PR is the idea of showing using the examples how to combine two services. We decided to leave that out of the SDK and have examples instead.

examples/conversation_addons/tone_detection.js
aprilwebster: Updated.

examples/conversation_addons/tone_detection.js
aprilwebster: Updated.

.project
aprilwebster: removed

examples/conversation_addons/tone_detection.js
nfriedly: @germanattanasio ESLint can both enforce and automatically fix this.. I should turn that rule on.

nfriedly: Although, in this case, it should just be `0.5`, no `parseFloat()` or strings. All numbers in JS are floats, so there's no need to do anything special here.

examples/tone_conversation_integration.v1.js
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * Instantiate the Watson Conversation Service
+ */
+var conversation = watson.conversation({
+  url: 'https://gateway.watsonplatform.net/conversation/api',
+  username: process.env.CONVERSATION_USERNAME || '<conversation_username>',
+  password: process.env.CONVERSATION_PASSWORD || '<conversation_password>',
+  version_date: '2016-07-11',
+  version: 'v1'
+});
```
nfriedly: FYI, you don't need to change anything here, but I made a updates to the SDK with v2 such that you can now create instances via `new watson.ConversationV1({/*...*/})`, and/or require the ConversationV1 file directly. 

I also changed it to automatically check those exact env properties if no credentials were passed in.

But, it's fully backwards compatible with everything here, so there's no problem with this example. (And explicitly calling out the env properties is probably good in an example.) I just thought I'd call it out since I'm going through and updating the docs and other examples this week.

nfriedly:
Great work @aprilwebster - I haven't tried it locally yet, but looking through the code it looks pretty sharp. 

There are a couple of little things called out here and by the [automated tests](https://travis-ci.org/watson-developer-cloud/node-sdk/jobs/151081119#L759) (you can run `npm test` to run those locally), and then this looks good to merge.

examples/tone_conversation_integration.v1.js
aprilwebster: Thanks for the heads up. :)

aprilwebster:
Thanks @nfriedly!  There seems to be an issue with the es6 fat arrows.  When I specify the es6 env inside the files, the automated tests pass.  Any suggestions?

nfriedly:
Oh, that's because the library officially supports node v0.12, 4.x, and 6.x right now, so the linter doesn't allow fat arrows. We'll probably drop v0.12 support in the next major release, so I think enabling it for that file is OK. (I considered dropping it for v2 even but decided not to...)

aprilwebster:
Thanks!  Enabled es6 and pushed changes.  :)

nfriedly:
Hey @aprilwebster, could you take a look at #300 (I was thinking I would just push it to this PR, but of course, I can't push code to your fork ;)

I just took this and did a little bit of reorganization mostly, there's details in the ticket.

PR #299, opened 2016-08-11
Targeted emotion

g-may:
Adding targeted emotion in the same style as targeted sentiment


alchemy-language/v1.js
```diff
@@ -16,7 +16,7 @@
 
 'use strict';
 
-var extend         = require('extend');
+var extend = require('extend');
```
nfriedly: :)

nfriedly:
It looks like the tests are failing - something to do with the exclamation point, maybe?

nfriedly:
Cool, thanks!

PR #304, opened 2016-08-18
Dialog deprecation

germanattanasio:


dialog/v1.js
```diff
@@ -31,6 +31,8 @@ var BaseService = require('../lib/base_service');
  */
 function DialogV1(options) {
   BaseService.call(this, options);
+  //eslint-disable-next-line no-console
+  console.warn('WARNING: The Dialog service was deprecated on August 15, 2016, existing instances of the service will continue to function until August 9, 2017. Users of the Dialog service should migrate their applications to use the Conversation service.');
```
nfriedly: We should let `options.silent` kill this message.  Especially with a year long transition period.

And you might want to just put https://www.ibm.com/watson/developercloud/doc/conversation/migration.shtml in instead of the second sentence.

germanattanasio:
```none
WARNING: The Dialog service was deprecated, existing instances of the service will continue to function until August 9, 2017. 
See https://www.ibm.com/watson/developercloud/doc/conversation/migration.shtml
Set {silent: true} to disable this message.
```

nfriedly:
Excellent!

PR #305, opened 2016-08-26
User can choose whether to maintain tone history for each utterance 

vibhasinghal:
### Summary

Depending on use-case, there might be a need to maintain tone history for each user utterance or not.  The earlier code version, by default always stored tone history. This update adds a flag "maintainToneHistoryInContext". Developer can set this value to false to not maintain tone history in context. 

When maintainToneHistoryInContext == true
Conversation context will have - tone for current utterance as well as tones for all prior user utterances.
When maintainToneHistoryInContext == false
Conversation context will only have - tone for current utterance

### DIFF
--- a/examples/browserify/README.md
+++ b/examples/browserify/README.md
@@ -17,11 +17,14 @@ The following services support CORS
 
  * Tradeoff Analytics
  * Tone Analyzer
- * Speech to Text
- * Text to Speech
+ * Speech to Text*
+ * Text to Speech*
  * Personality Insights
  * Document Conversion
  * All Alchemy services
+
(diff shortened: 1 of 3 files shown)


nfriedly:
Hi @vibhasinghal,

This seems to add a fair amount of complexity. I understand that this could be important in a production app, but do you feel that it is necessary in example code?

vibhasinghal:
Yes. The idea behind example code is to enable developers to have code they could quickly reuse. 
The need to not save tone history is something we encountered in just the first tone-app we started creating. So I guess many developers would need it. 

In terms of code, it is not adding complexity. There are just a bunch of if checks in 3 places and an extra variable being passed.

nfriedly:
Alright, I'll trust you on this one.

vibhasinghal:
Merci !

PR #312, opened 2016-09-17
fixed an error in the Conversation service example

knoid:
A very simple fix for the conversation example in the readme file, it was missing the `text` key to work properly.


germanattanasio:
thanks @knoid

Issue #318, opened 2016-09-27
handle language translator / language translation differences


nfriedly:
The new language translation endpoint is now live (`https://gateway.watsonplatform.net/language-translator/api/v2/`), and while the API for the two endpoints is the same, they are billed differently and do not accept each others credentials :(

The SDK already migrated to the new naming, but using the old endpoint and assuming that it would be a seamless change. Since it isn't we need to do some work. At a minimum:
- Re-Create the `LanguageTranslaionV2` class for the old endpoint
- Update `LanguageTranslatorV2` to use the new endpoint

However, this on it's own will cause pain for most people who are currently using the SDK. So, it either needs to be a breaking change (SDK v3) or else needs to have some extra magic under the hood, such as testing the given credentials against each endpoint to see which one they match.

nfriedly:
Update: The SDK now has both Language Translation and Translator classes, both default to the old url, and the new Translator class logs a warning if you do not specify a url. The next major release, hopefully to be delivered within a few weeks, will change the default url for Translator.

PR #319, opened 2016-09-27
clarify language translator vs translation

nfriedly:
Just update the readme with instructions for how to use the current sdk with either service


nfriedly:
This is a band-aid fix for #318

PR #329, opened 2016-10-25
STT customization support

nfriedly:
I tweaked the handling of requiredparams a bit so that you can pass the original params to requestwrapper and then it'll use that instead of trying to rebuild it for the requiredparams check.

And then I implemented all of the STT customization API + a helper to notify you when processing / training is complete. And integration tests for everything. 

I should probably do unit tests too...

### DIFF
--- a/speech-to-text/v1.js
+++ b/speech-to-text/v1.js
@@ -770,8 +770,8 @@ SpeechToTextV1.ERR_NO_CORPORA = 'ERR_NO_CORPORA';
  *
  * @param {Object} params
  * @param {String} params.customization_id
- * @param {Number} [params.interval=1000] - (milliseconds) - how log to wait between status checks
- * @param {Number} [params.times=150] - maximum number of attempts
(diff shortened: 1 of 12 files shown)


speech-to-text/v1.js
```diff
@@ -371,4 +373,664 @@ SpeechToTextV1.prototype.createRecognizeStream = function(params) {
...
+          return next(err)
+        }
+        next();
+      })
+    },
+
+    // check the customization status repeatedly until it's ready or avaliable
+    function(next) {
+      var options = extend({
+        interval: 1000,
```
germanattanasio: I think we should check every 5 seconds.

kognate: Could it take the interval as a function parameter with a default value?

nfriedly: @kognate: It does allow interval and times to be overridden. 

@german: I originally wrote this for the integration tests, where 1 second makes a little more sense (because it's uploading tiny files), but I think you're right that the general case should probably be higher.

PR #331, opened 2016-11-07
initial support for discovery

germanattanasio:


package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "watson-developer-cloud",
+  "version": "2.6.1",
```
kognate: I'm not sure that two package versions is legit.

germanattanasio: hahahaha

PR #334, opened 2016-11-11
removes setting of charset on html content type

### LINKED ISSUE #332: [document-conversion] Incorrectly forces UTF-8 for HTML
In the current `master` branch of the SDK, the `DocumentConversionV1` incorrectly forces the `UTF-8` charset on HTML documents.  See https://github.com/watson-developer-cloud/node-sdk/blob/master/document-conversion/v1.js#L70

It should do one or both of the following:
1. Allow the user to specify the charset `X` of the input document and set the file part's `contentType` to `text/html; charset=X`.
2. Default the file part's `contentType` to `text/html` without specifying the `charset`.

The assertion that the service [requires utf-8](https://github.com/watson-developer-cloud/node-sdk/blob/master/document-conversion/v1.js#L66) is incorrect: If the charset is not specified in the file part, the service will look for it in the HTML's `meta` tags.

kognate:
it has been reported that setting the charset is not required
for html files, this removes the auto-setting of charset for html files
content types.

Fixes #332

### DIFF
--- a/document-conversion/v1.js
+++ b/document-conversion/v1.js
@@ -63,11 +63,10 @@ function fixupContentType(params) {
     };
   }
   else if (params.file.path && /.html?$/.test(params.file.path)) {
-    // for HTML, the service requires that a utf-8 charset be specified in the content-type
     params.file = {
       value: params.file,
       options: {
-        contentType: 'text/html; charset=utf-8'
+        contentType: 'text/html'
       }
     };
   }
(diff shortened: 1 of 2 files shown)


nfriedly:
Oh, good, they fixed it!

nfriedly:
Op, we should really get rid of the [`fixupContentType()`](https://github.com/watson-developer-cloud/node-sdk/blob/master/document-conversion/v1.js#L56) function entirely, I'll do that in a minute.

nfriedly:
Actually, part of that is still needed, it's just the special case for html - request can handle it automatically now that the service accepts it without `charset=8tf-8` in the content-type.

PR #342, opened 2016-12-01
Updated to add stubs for basic discovery commands

kognate:
Discovery commands of collections, environments, and query are in now.

### DIFF
--- a/test/integration/test.speech_to_text.js
+++ b/test/integration/test.speech_to_text.js
@@ -102,7 +102,10 @@ describe('speech_to_text_integration', function() {
       assert(res.results);
       assert(res.results[0]);
       assert(res.results[0].keywords_result);
-      assert.deepEqual(Object.keys(res.results[0].keywords_result), ['tornadoes','hail', 'rain']);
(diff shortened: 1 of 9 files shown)


germanattanasio:
I made some changes to make it be `method(params,callback)`

test/integration/test.speech_to_text.js
```diff
@@ -77,14 +77,14 @@ describe('speech_to_text_integration', function() {
                 {
                   "normalized_text": "rain",
                   "start_time": 5.59,
-                  "confidence": 0.986,
+                  "confidence": 0.994,
```
nfriedly: I think this is flakeyness at the service level.. because other tests are passing with the old values. I think we are going to have to either loosen up this test or else skip it entirely for now.

nfriedly:
I added a couple of things (readme, jsdoc, version_date constant). I'm also going to loosen up that STT test so that it stops flaking out on us

nfriedly:
Should be released in `v2.11.0` assuming no more services flake out and kill our CI loop...

PR #350, opened 2016-12-07
expanded discovery calls

kognate:
Adds ability to create collections and add documents

### DIFF
--- a/discovery/v1.js
+++ b/discovery/v1.js
@@ -31,7 +31,7 @@ function DiscoveryV1(options) {
 
   // Check if 'version_date' was provided
   if (typeof this._options.version_date === 'undefined') {
-    throw new Error('Argument error: version_date was not specified, use DiscoveryV1.VERSION_DATE_2016_07_11');
+    throw new Error('Argument error: version_date was not specified, use DiscoveryV1.VERSION_DATE_2016_12_15');
   }
   this._options.qs.version = options.version_date;
 }
@@ -46,7 +46,7 @@ DiscoveryV1.URL = 'https://gateway.watsonplatform.net/discovery-experimental/api
  * Initial release
  * @type {string}
  */
-DiscoveryV1.VERSION_DATE_2016_07_11 = '2016-11-07';
+DiscoveryV1.VERSION_DATE_2016_12_15 = '2016-12-15';
 
 
 
--- a/discovery/v1.js
+++ b/discovery/v1.js
@@ -38,8 +38,8 @@ function DiscoveryV1(options) {
 
 util.inherits(DiscoveryV1, BaseService);
 DiscoveryV1.prototype.name = 'discovery';
-DiscoveryV1.prototype.version = 'v1-experimental';
-DiscoveryV1.URL = 'https://gateway.watsonplatform.net/discovery-experimental/api';
+DiscoveryV1.prototype.version = 'v1';
+DiscoveryV1.URL = 'https://gateway.watsonplatform.net/discovery/api';
 
 
 /**
--- a/discovery/v1.js
+++ b/discovery/v1.js
@@ -1,0 +1,331 @@
+/**
+ * Copyright 2015 IBM Corp. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+var util = require('util');
+var requestFactory = require('../lib/requestwrapper');
+var BaseService = require('../lib/base_service');
+var pick = require('object.pick');
+
+/**
+ *
+ * @param options
+ * @constructor
+ */
+function DiscoveryV1(options) {
+  BaseService.call(this, options);
+
+  // Check if 'version_date' was provided
+  if (typeof this._options.version_date === 'undefined') {
+    throw new Error('Argument error: version_date was not specified, use DiscoveryV1.VERSION_DATE_2016_07_11');
+  }
+  this._options.qs.version = options.version_date;
+}
+
+util.inherits(DiscoveryV1, BaseService);
+DiscoveryV1.prototype.name = 'discovery';
+DiscoveryV1.prototype.version = 'v1-experimental';
+DiscoveryV1.URL = 'https://gateway.watsonplatform.net/discovery-experimental/api';
+
+
+/**
+ * Initial release
+ * @type {string}
+ */
+DiscoveryV1.VERSION_DATE_2016_07_11 = '2016-11-07';
+
+
+
+/**
+ * Return the list of environments
+ *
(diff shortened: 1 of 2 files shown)


test/unit/test.discovery.v1.js
```diff
@@ -0,0 +1,177 @@
...
+    add_document: '/v1/environments/env-guid/collections/col-guid/documents',
+    delete_document: '/v1/environments/env-guid/collections/col-guid/documents/document-guid',
+    query: '/v1/environments/env-guid/collections/col-guid/query'
+  };
+
+  before(function() {
+    nock.disableNetConnect();
+    nock(service.url)
+      .persist()
+      .post(paths.environments + '?version=' + service.version_date)
```
kognate: Yeah, I copied this from the v1-experimental tests.

discovery/v1.js
```diff
@@ -0,0 +1,331 @@
...
+DiscoveryV1.prototype.name = 'discovery';
+DiscoveryV1.prototype.version = 'v1-experimental';
+DiscoveryV1.URL = 'https://gateway.watsonplatform.net/discovery-experimental/api';
+
+
+/**
+ * Initial release
+ * @type {string}
+ */
+DiscoveryV1.VERSION_DATE_2016_07_11 = '2016-11-07';
```
nfriedly: Is this the correct version_date for the GA release?

discovery/v1.js
```diff
@@ -0,0 +1,331 @@
...
+  };
+  return requestFactory(parameters, callback);
+};
+
+/**
+ * Return the list of collections in the given environment
+ *
+ * @param {Object} params
+ * @param {String} params.environment_id
+ * @param {String} [params.name] Find collections with the given name.
```
nfriedly: I don't see params.name used anywhere in this method.

(It looks like this is also wrong in https://github.com/watson-developer-cloud/node-sdk/blob/master/discovery/v1-experimental.js#L103 )

discovery/v1.js
```diff
@@ -0,0 +1,331 @@
...
+  };
+  return requestFactory(parameters, callback);
+};
+
+/**
+ * Add a document to a collection
+ * @param params
+ * @param {String} params.environment_id environment guid for the collection
+ * @param {string} params.collection_id the guid of the collection to delete
+ * @param {file} params.file a file to post (smaller than 50mb)
```
nfriedly: I think this should be `{Buffer|ReadableStream|Object}` rather than `{file}` (?)

(Object allows you to specify filename and/or content type, e.g. `{value: Buffer, options: contentType: 'text/foo'}`.. or something like that.)

discovery/v1.js
```diff
@@ -0,0 +1,331 @@
...
+ * @param {string} [params.configuration_id] config guid
+ * @param {string} [params.metadata] file metadata, including content-type (will infer if missing)
```
nfriedly: String? what should this look like?

discovery/v1.js
```diff
@@ -0,0 +1,331 @@
...
+  // Check if 'version_date' was provided
+  if (typeof this._options.version_date === 'undefined') {
+    throw new Error('Argument error: version_date was not specified, use DiscoveryV1.VERSION_DATE_2016_07_11');
+  }
+  this._options.qs.version = options.version_date;
+}
+
+util.inherits(DiscoveryV1, BaseService);
+DiscoveryV1.prototype.name = 'discovery';
+DiscoveryV1.prototype.version = 'v1-experimental';
```
nfriedly: This should just be `v1`

discovery/v1.js
```diff
@@ -0,0 +1,331 @@
...
+DiscoveryV1.URL = 'https://gateway.watsonplatform.net/discovery-experimental/api';
```
nfriedly: Presumably this shouldn't have `-experimental` either (?)

test/unit/test.discovery.v1.js
```diff
@@ -0,0 +1,177 @@
...
+
+  describe('discovery()', function() {
+    it('should generate a valid payload', function() {
+      var req = discovery.getEnvironments({}, noop);
+      assert.equal(req.uri.href, service.url + paths.environments + '?version=' + service.version_date);
+      assert.equal(req.method, 'GET');
+    });
+
+    it('should check no parameters provided (negative test)', function() {
+      discovery.getEnvironments({}, missingParameter);
```
nfriedly: This test doesn't make any sense. It passes regardless of weather or not getEnvironments() throws a missingParameter error. (Because the test is sync, but if getEnvironments doesn't throw an error, then fires the callback async - so the test has already passed by the time `missingParameter` throws). I think we should probably make the callback always async, and make `missingParameter` accept a callback.

test/unit/test.discovery.v1.js
```diff
@@ -0,0 +1,177 @@
...
+      var threw = false;
+      try {
+        // eslint-disable-next-line
+        var discovery = new Discovery(service1);
+      }
+      catch(err) {
+        threw = true;
+        assert.equal(err.message, 'Argument error: version_date was not specified, use DiscoveryV1.VERSION_DATE_2016_07_11');
+      }
+      assert(threw, 'should throw an error')
```
nfriedly: BTW, there is an `assert.throws` method ;)

https://nodejs.org/api/assert.html#assert_assert_throws_block_error_message

test/unit/test.discovery.v1.js
```diff
@@ -0,0 +1,177 @@
...
+      .post(paths.collections + '?version=' + service.version_date)
+      .reply(200, {'collection_id': 'yes'})
+      .delete(paths.delete_collection + '?version=' + service.version_date)
+      .reply(200, {'config': 'yes'})
+      .post(paths.add_document + '?version=' + service.version_date)
+      .reply(200, {'add_doc': 'yes'})
+      .delete(paths.delete_document + '?version=' + service.version_date)
+      .reply(200, {'delete_doc': 'yes'})
+      .get(paths.configurations + '?version=' + service.version_date)
+      .reply(200, {'configs': 'yes'});
```
nfriedly: FYI: these don't hurt, but nock can't check things in time for for sync tests because the test passes before nock gets called, and any error that nock might throw gets swallowed by mocha because it occurs outside of a test :/

discovery/v1.js
kognate: done.

discovery/v1.js
kognate: nope.

discovery/v1.js
kognate: changed

discovery/v1.js
kognate: fixed. and it's right in experimental.

discovery/v1.js
kognate: changed

discovery/v1.js
kognate: they should look at the docs.  at least isn't that what the other discussion concluded?

test/unit/test.discovery.v1.js
kognate: I've no idea what that was doing there.  removed.

test/unit/test.discovery.v1.js
kognate: good to know.  I copied that pattern from another test.  fixed.

discovery/v1.js
nfriedly: BTW, I'm not seeing the change in the PR - perhaps you put it in a different branch or forgot to push it to github?

kognate:
@nfriedly  I did miss a file.  it's in now.

nfriedly:
Should be released in `v2.14.0` shortly.

PR #353, opened 2016-12-08
Add function that waits until the status of STT corpus is analyzed

luthraprerna:
### Summary

Added a function  that waits until the status of STT corpus is analyzed

### DIFF
--- a/speech-to-text/v1.js
+++ b/speech-to-text/v1.js
@@ -841,9 +841,9 @@ SpeechToTextV1.prototype.whenCustomizationReady = function(params, callback) {
 //Check if there is a corpus that is still being processed
 function isProcessing(corporaList) {
   var recordsBeingProcessed = corporaList.corpora.filter(function(record) {
-    return record['status'] == 'being_processed';
+    return record['status'] === 'being_processed';
   });
-  if (recordsBeingProcessed.length == 0) {
+  if (recordsBeingProcessed.length === 0) {
     return false;
   } else {
     return true;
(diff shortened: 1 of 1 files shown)


speech-to-text/v1.js
```diff
@@ -837,6 +838,97 @@ SpeechToTextV1.prototype.whenCustomizationReady = function(params, callback) {
   });
 };
 
+//Check if there is a corpus that is still being processed
+function isProcessing(corporaList) {
+  var recordsBeingProcessed = corporaList.corpora.filter(function(record) {
+    return record['status'] == 'being_processed';
```
nfriedly: You must use `===` in order to make the linter happy. Run `npm test` locally to make sure everything will pass.

nfriedly:
Thanks, looks good overall, I just need two things from you:

* Replace `==` with `===` to make the tests pass - https://travis-ci.org/watson-developer-cloud/node-sdk/jobs/182310079#L776
* Complete the Contributor License Agreement: https://cla-assistant.io/watson-developer-cloud/node-sdk?pullRequest=353 (Doesn't really matter if you're an IBMer, but it does make things easier and only takes a few seconds.)

Also, in the future, please separate the formatting and code changes into separate commits, but I don't think that needs to block this merge.

luthraprerna:
Thanks for reviewing the PR. I've replaced == with ===. I ran `npm test` to make sure all tests are passed. Please look at the most recent commit.

germanattanasio:
@luthraprerna can you add an example in the `examples` folder showing how to use this feature? we also have a CHANGELOG where we list the changes we made.

nfriedly:
I usually take care of the changelog (since it's updated for each release).

nfriedly:
Should be released in `v2.13.0` once the CI loop finishes. I wouldn't be against an example if you want to send another PR, but as this is just a compliment of the existing `whenCustomizationReady()` helper method, I wouldn't consider that a blocker.

luthraprerna:
Thanks @nfriedly for merging the PR. I've created a separate PR whereby I've added an example.

nfriedly:
Cool, thanks!

PR #362, opened 2016-12-15
Fix content type for language translator

germanattanasio:


nfriedly:
Released in v2.14.2 - `npm install --save watson-developer-cloud@latest`

PR #370, opened 2017-01-05
Fixes issue where adding a document as buffer/string fails

### LINKED ISSUE #369: [discovery] document additions fail
This was reported internally and on SO.  https://stackoverflow.com/questions/41473388/create-new-document-in-ibm-watson-discovery-service-with-watson-developer-cloud

There is a bug in the discovery add document code.

kognate:
The problem was that the filename wasn't being set
correctly in the multipart form upload and the service was rejecting
it.  This also adds an example for discovery that shows adding a document.

Fixes #369

### DIFF
--- a/discovery/v1.js
+++ b/discovery/v1.js
@@ -259,9 +259,16 @@ DiscoveryV1.prototype.addDocument = function(params, callback) {
 
   // if we get a buffer or object, we need to include stuff about filename for the service
   if (formDataParams.file) {
-    if (typeof formDataParams.file.filename !== 'string') {
+    if (typeof formDataParams.file.filename !== 'string' &&
+          !(formDataParams.file.options &&
+             typeof formDataParams.file.options.filename !== 'string') &&
+          !(formDataParams.file.path &&
+            typeof formDataParams.file.path !== 'string') &&
+          !(formDataParams.file.name &&
+            typeof formDataParams.file.name !== 'string')) {
       var filedat = formDataParams.file
-      formDataParams.file = { value: filedat, options: { filename: 'inferedname'}};
+      // the filename used below is because the name must exist
+      formDataParams.file = { value: filedat, options: { filename: '_'}};
     }
   }
 
@@ -315,7 +322,7 @@ DiscoveryV1.prototype.deleteDocument = function(params, callback) {
  * @param {String} [params.query]  A query search returns all possible results, even when it's not very relevant, with the most relevant documents listed first. Use a query search when you want to find the most relevant search results. Results are scored between 0 and 1, with 1 being an exact match and 0 being not a match at all.
  * @param {String} [params.aggregation] An aggregation search uses combinations of filters and query search to return an exact answer. Aggregations are useful for building applications, because you can use them to build lists, tables, and time series. For a full list of possible aggregrations, see the Query reference.
  * @param {Number} [params.count=10] Number of documents to return
- * @param {String} [params.return] A comma sepkarated list of the portion of the document hierarchy to return.
+ * @param {String} [params.return] A comma separated list of the portion of the document hierarchy to return.
  * @param {Number} [params.offset=0] For pagination purposes. Returns additional pages of results. Deep pagination is highly unperformant, and should be avoided.
  */
 DiscoveryV1.prototype.query = function(params, callback) {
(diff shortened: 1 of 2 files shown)


examples/discovery.v1.js
```diff
@@ -0,0 +1,29 @@
...
+discovery.getEnvironments({}, function(error, data) {
+  console.log(JSON.stringify(data, null ,2));
+});
+
+var buffer = fs.readFileSync('../test/resources/sampleHtml.html');
+
+discovery.addDocument({
+  environment_id: '3526dd92-5b18-4284-b5c0-f20e2f92307f',
+  collection_id: '90203b57-3de9-4db8-b11b-2a353ad99a4d',
+  file: buffer,
```
nfriedly: It's good that buffers work, but I'd recommend that the primary example use fs.createReadStream() (same result, less memory usage)

discovery/v1.js
```diff
@@ -306,7 +315,7 @@ DiscoveryV1.prototype.deleteDocument = function(params, callback) {
  * @param {String} [params.query]  A query search returns all possible results, even when it's not very relevant, with the most relevant documents listed first. Use a query search when you want to find the most relevant search results. Results are scored between 0 and 1, with 1 being an exact match and 0 being not a match at all.
  * @param {String} [params.aggregation] An aggregation search uses combinations of filters and query search to return an exact answer. Aggregations are useful for building applications, because you can use them to build lists, tables, and time series. For a full list of possible aggregrations, see the Query reference.
  * @param {Number} [params.count=10] Number of documents to return
- * @param {String} [params.return] A comma separated list of the portion of the document hierarchy to return.
+ * @param {String} [params.return] A comma sepkarated list of the portion of the document hierarchy to return.
```
kognate: fixed. doh.

discovery/v1.js
```diff
@@ -255,14 +255,23 @@ DiscoveryV1.prototype.addDocument = function(params, callback) {
   params = params || {};
 
   var query_params = pick(params, ['configuration_id'])
+  var formDataParams = pick(params, ['file', 'metadata']);
+
+  // if we get a buffer or object, we need to include stuff about filename for the service
+  if (formDataParams.file) {
+    if (typeof formDataParams.file.filename !== 'string') {
```
kognate: fixed.

The filename can be inferred from file streams, but not if they're a buffer.

discovery/v1.js
```diff
@@ -255,14 +255,23 @@ DiscoveryV1.prototype.addDocument = function(params, callback) {
...
+      var filedat = formDataParams.file
+      formDataParams.file = { value: filedat, options: { filename: 'inferedname'}};
```
kognate: fixed.

examples/discovery.v1.js
kognate: fixed.  This is the integration test,  FWIW.  The service is really hard to write automation tests for because we're sharing a key.  We need to get a 'test' instance at some point.

discovery/v1.js
```diff
@@ -255,14 +255,30 @@ DiscoveryV1.prototype.addDocument = function(params, callback) {
   params = params || {};
 
   var query_params = pick(params, ['configuration_id'])
+  var formDataParams = pick(params, ['file', 'metadata']);
+
+  // if we get a buffer or object, we need to include stuff about filename for the service
+  if (formDataParams.file) {
+    if (typeof formDataParams.file.filename !== 'string' &&
```
nfriedly: Does file.filename work?

discovery/v1.js
```diff
@@ -255,14 +255,30 @@ DiscoveryV1.prototype.addDocument = function(params, callback) {
...
+          !(formDataParams.file.options &&
+             typeof formDataParams.file.options.filename !== 'string') &&
+          !(formDataParams.file.path &&
+            typeof formDataParams.file.path !== 'string') &&
+          !(formDataParams.file.name &&
+            typeof formDataParams.file.name !== 'string')) {
```
nfriedly: Man, I feel like I've helped to create a monstrosity.... 

We could probably just check for truthyness and not worry about the typeof check - .toString() works on prettymuch everything. That would save us a couple of lines here.

nfriedly:
Merging, not sure what's happening on travis

Issue #368, opened 2017-01-02
[text-to-speech] No way of handling errors correctly before a pipe to the FS


deanc:
# Steps to reproduce

Pass in some bad params to the `synthesize` function e.g.
```
    var params = {
        text: "Hello world",
        voice: "asdf", // bad voice...
        accept: 'audio/ogg'
    };
```

# Expected behavior:

Should be able to catch some kind of error before piping. For example:

```
text_to_speech.synthesize(params)
        .on("error", function (err) {
            console.log(err); // this should prevent the pipe from ever being reached
        })
        .pipe(fs.createWriteStream(destinationPath + "/" + hash + '.ogg'));
```
# Actual behavior
The error is never emitted, and the file is always written to the FS.

# Node version
7.3.0

# SDK version
```"watson-developer-cloud": "^2.14.4"```

# Summary

As it stands errors are sent as JSON, meaning the error itself is written to the file system. This is not ideal. Please advise.

**Comment by nfriedly** [comment]
Hi @deanc 

I agree that this is not ideal behavior for the streaming API. The SDK currently uses [request](https://www.npmjs.com/package/request) under the hood which suffers from the same issue. 

That said, your desired behavior is possible with a couple of tweaks to your code:

```js
var stream = textToSpeech.synthesize({
  text: 'Hello from IBM Watson',
  voice: 'INVALID_VOICE', // comment this line to show the example working correctly
  accept: 'audio/wav' // default is audio/ogg; codec=opus
}, function(error) {
  if (error) {
    console.log(error);
  }
})
.on('response', function(res) {
  if (res.statusCode === 200) {
    stream.pipe(fs.createWriteStream('output.wav'));
  }
  // errors will be handled by the callback after the body is downloaded
});
```

This works by hooking into the `response` event that request provides. This event is always emitted before any data, and thus is a safe way to check the status code and headers before piping the output. 

I also put a copy of this in https://github.com/watson-developer-cloud/node-sdk/blob/master/examples/text_to_speech_synthesize_handle_errors.js

I'll also look into improving the streaming API to emit error events instead of streaming the data for non-200 statuses, but I'm not sure if/when that would be implemented.

deanc:
Hi @nfriedly . Thanks for the reply!

I ended up going down the route of checking the response status code as well as the returned mime type. So this is line with your proposed solution.

[...]

PR #377, opened 2017-01-17
Returns a node stream if callback is null

germanattanasio:
### Summary

Returns a node stream and emit an error if the callback is null. Useful when using streams instead of callbacks:

```js
textToSpeech.synthesize({})
  .on('error', function(error) {
     console.log(error); // output: "Error: Missing required parameters: text"
  })
  .pipe(fs.createWriteStream('../resources/tts-output.ogg'));
```

I added a test case and also updated STT and TTS to use `requiredParams` instead of manually checking for missing parameters.

### DIFF
--- a/lib/requestwrapper.js
+++ b/lib/requestwrapper.js
@@ -21,7 +21,7 @@ var request     = require('request');
 var pkg         = require('../package.json');
 var helper      = require('./helper');
 var parseString = require('string');
-
+var readableStream = require('stream').PassThrough;
 var isBrowser = typeof window === "object";
 
 function parsePath(path, params) {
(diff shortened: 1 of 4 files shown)


germanattanasio:
There is a test failing for tradeoff analytics :(

nfriedly:
Not sure what's up with the failing TA test.. yesterday and today it's been consistently failing with a zlib unexpected-end error, but only on node v6 & 7 on travis. It always passes on node v4 on travis and any version of node I've tested locally. I may disable the test if it keeps up.

nfriedly:
Fixes #368

PR #387, opened 2017-02-01
Adds support for NLU

kognate:
NLU has support for the analyze method.


natural-language-understanding/v1.js
```diff
@@ -0,0 +1,184 @@
...
+ * @param options
+ * @constructor
+ */
+function NaturalLanguageUnderstandingV1(options) {
+  BaseService.call(this, options);
+}
+util.inherits(NaturalLanguageUnderstandingV1, BaseService);
+NaturalLanguageUnderstandingV1.prototype.name = 'natural_language_understanding';
+NaturalLanguageUnderstandingV1.prototype.version = 'v1';
+NaturalLanguageUnderstandingV1.URL = 'https://gateway-s.watsonplatform.net/natural-language-understanding/api';
```
germanattanasio: remove `-s` from `gateway-s`

natural-language-understanding/v1.js
```diff
@@ -0,0 +1,184 @@
...
+};
+
+/**
+  * Analyze the query.
+  * @params {QueryBuilder} query a QueryBuilder query
+  * @params {object} params any addtional request params, can be null
+  * @params {function} callback taking (error,  jsonResult)
+  * @returns {void}
+  */
+NaturalLanguageUnderstandingV1.prototype.analyze = function(query, params, callback) {
```
germanattanasio: I think we want to keep the signature to be `method(params, callback)`. It's a mental model at this point, having an extra parameter will make it difficult to differentiate when something is a query or when it's a param.

natural-language-understanding/v1.js
```diff
@@ -0,0 +1,184 @@
...
+var util = require('util');
+var BaseService = require('../lib/base_service');
+
+/**
+ * NaturalLanguageUnderstanding
+ * @param options
+ * @constructor
+ */
+function NaturalLanguageUnderstandingV1(options) {
+  BaseService.call(this, options);
```
germanattanasio: We should throw an error if `version_date` is not specified. This is something we do in other services that have versioning. We don't want to use a default version date because defaults may change and it could lead to users getting different responses if they use different versions of the SDK

test/unit/test.natural_language_understanding.v1.js
```diff
@@ -0,0 +1,39 @@
...
+    nock(watson.NaturalLanguageUnderstandingV1.URL)
+      .persist()
+      .post(`/v1/analyze?version=${watson.NaturalLanguageUnderstandingV1.VERSION_DATE}`)
+      .reply(200, {});
+
+    const query = new nlu.QueryBuilder();
+    assert(typeof(query._options.features) === 'undefined');
+    query.getAllFeatures().withTextString('hello this is a test');
+    assert(Object.keys(query._options.features).length > 0);
+    nlu.analyze(query, null, done);
```
germanattanasio: The use of `null` is what we are trying to avoid using `method(params,callback)`. In this case, you don't use `params` so it's null.

natural-language-understanding/v1.js
kognate: it won't work if we remove the '-s' for now.

natural-language-understanding/v1.js
kognate: ok, it throws now if the version is not included.

natural-language-understanding/v1.js
germanattanasio: You can use `NATURAL_LANGUAGE_UNDERSTANDING_URL`

kognate: the `.URL` var is special and is used by the base service.  If I change it to NATURAL_LANGUAGE_UNDERSTANDING_URL I'll have to modify the base service.

natural-language-understanding/v1.js
```diff
@@ -0,0 +1,82 @@
...
+  * @param {object} [params.features.keywords] - keywords feature
+  * @param {object} [params.features.categories] - categories feature
+  * @param {object} [params.features.emotion] - emotion feature
+  * @param {object} [params.features.sentiment] - sentiment feature
+  * @param {object} [params.features.relations] - relations feature
+  * @param {object} [params.features.semantic_roles] - semantic roles feature
+  * @params {function} callback taking (error,  jsonResult)
+  * @example
+  * const options = { 'text': 'I am some text to analyze, am I not cool?',
+  *                   'features': {
```
nfriedly: BTW, you can use markdown here (tripple-backtick-js) and it will be rendered properly in the docs :)

natural-language-understanding/v1.js
```diff
@@ -0,0 +1,82 @@
...
+  if (typeof this._options.version_date === 'undefined') {
+    throw new Error('Argument error: version_date was not specified, use 2016-01-23');
+  }
+  this._options.qs.version = this._options.version_date;
+}
+util.inherits(NaturalLanguageUnderstandingV1, BaseService);
+NaturalLanguageUnderstandingV1.prototype.name = 'natural_language_understanding';
+NaturalLanguageUnderstandingV1.prototype.version = 'v1';
+NaturalLanguageUnderstandingV1.URL = 'https://gateway.watsonplatform.net/natural-language-understanding/api';
+NaturalLanguageUnderstandingV1.VERSION_DATE = '2016-01-23';
```
nfriedly: Op, github apparently lost this comment in the original review, but I had said that this should be named

`NaturalLanguageUnderstandingV1.VERSION_DATE_2016_01_23`

Otherwise it kind of defeats the purpose of having a version date.

test/resources/natural_language_classifier/energy-policy.html
```diff
@@ -0,0 +1,535 @@
+<!DOCTYPE html>
```
kognate: this is just a typo. Fixing.

.eslintrc.js
```diff
@@ -1,6 +1,14 @@
...
     "env": {
-        "node": true
+        "node": true,
+	es6: true
+    },
+    "parserOptions": {
+        "ecmaVersion": 6,
+   	"sourceType": "module",
+        "ecmaFeatures": {
+            "jsx": true
```
kognate: ok.

PR #395, opened 2017-02-13
Allow to update existing document in addDocument function

pudelduscher:
Allow to update existing documents in addDocument function for the discovery service like its described in the documentation


chughts:
I think that this should be an updateDocument method.

germanattanasio:
@pudelduscher can you add the `updateDocument` instead of modifying `addDocumented` ?

pudelduscher:
i will close this request and switch the branch

PR #399, opened 2017-02-17
display serviceName that generate an error in 'getMissingParams'

adrien2p:
Add method in the `helper.j`s file to get the name of the original caller that generate an error catch by "getMissingParams".

### DIFF
--- a/lib/helper.js
+++ b/lib/helper.js
@@ -44,7 +44,7 @@ module.exports = {
       });
     }
 
-    return missing.length > 0 ? new Error('Missing required parameters: ' + missing.join(', ')) : null;
+    return missing.length > 0 ? new Error('Missing required parameters: ' + missing.join(', ') + ' (' + this.getServiceNameFromStackTrace() + ')') : null;
   },
   /**
    * Return true if 'text' is html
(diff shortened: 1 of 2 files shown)


germanattanasio:
This looks like a hack, @nfriedly  what do you think?

nfriedly:
This seems a bit hackey, but I get that it does solve a legitimate issue. Let me poke at it a bit and at least make sure it doesn't do anything gnarly in browsers.

lib/helper.js
```diff
@@ -71,5 +71,18 @@ module.exports = {
...
+   * @returns {*}
+   */
+  getServiceNameFromStackTrace: function() {
+    const stack = new Error().stack.toString();
+    const stackArr = stack.split(' at ');
+    const callers = stackArr.filter(trace => {
+      return /node-sdk/.test(trace) && !/node_modules/.test(trace) && !/anonymous/.test(trace) && trace;
+    });
+
+    return callers[callers.length - 1].split(' (')[0];
```
adrien2p: @nfriendly you completely right, i think the refactor as you talked is the best solution. It's a bigger changes as i thought. All other easier solutions are non standard or obsolete, so i think the best is waiting for refactor and close this PR, what do you think ?

adrien2p: I think the async callback stt is prioritu

nfriedly: @adrien2p Yea, I think you're right. I'll kill this one for now and try to find time to roll out a better version of this + some of the other improvements that are currently sitting on the back burner.

PR #401, opened 2017-02-19
[Discovery] Allow to update existings documents

pudelduscher:


pudelduscher:
@germanattanasio i hope this is what you want

germanattanasio:
This looks good. One more thing, can you do me a favor and update `query_params` to be `queryParams`? I know it wasn't you but since you are modifying the file.

pudelduscher:
@germanattanasio should i update it everywhere or just in the update method?

germanattanasio:
In every method withing the discovery file.

PR #406, opened 2017-02-23
conversation example with two messages

boazdavid:
##### Checklist

- [x] tests are included
- [x] link to public docs when adding new a service or new features for an existing service

### DIFF
--- a/examples/conversation.v1.js
+++ b/examples/conversation.v1.js
@@ -25,6 +25,7 @@ const message = function(text) {
     input: {
       text: text
     },
+    // pass the context from previous response
     context: currentContext
   };
   return new Promise((resolve, reject) => conversation.message(payload, function(err, data) {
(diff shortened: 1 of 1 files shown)


examples/conversation.v1.js
```diff
@@ -13,24 +13,51 @@ const conversation = new watson.ConversationV1({
...
-  }
+// This example makes two successive calls to conversation service.
+message('first message').then(response1 => {
+  // APPLICATION-SPECIFIC CODE TO PROCESS THE DATA
+  // FROM CONVERSATION SERVICE
+  console.log(JSON.stringify(response1, null, 2));
+
+  // invoke a second call to conversation
+  message('second message')
+    .then(response2 => {
```
nfriedly: I think you should return `message('second message')` here and move the `.then()` and `.catch()` to the outside. That way the `.catch()` handle errors from either message.

examples/conversation.v1.js
```diff
@@ -13,24 +13,51 @@ const conversation = new watson.ConversationV1({
...
+ * Conversation service is stateless.
+ * The application need to maintain the conversation state.
+ * A context object is returned in the messgage response and the app should pass it in the next message request
  */
-const payload = {
-  workspace_id: process.env.WORKSPACE_ID || '<workspace_id>',
-  input: {
-    text: 'I am happy to talk to you today.'
-  }
+let currentContext = undefined;
```
nfriedly: I think this should be a second parameter to the `message()` function, rather than a global variable. Then, below where you call `message('second message')`, you could change it to `message('second message', response1.context)`. 

I think that would make the example easier for users to build on.

boazdavid:
Thanks, please review

PR #408, opened 2017-02-27
[Conversation] Update SDK with news APIs

Themandunord:
Hi,

Add new API for Conversation : 

Get intents
Create intent
Delete intent
Get intent
Update intent

With tests.

It's my first pull request, don't hesitate to tell me if it's correct or not :)

Rémy.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,6 @@
+# https://git-scm.com/book/en/v2/Customizing-Git-Git-Attributes
+
+# Code linting require line-endings be LF rather than CRLF in js files. This file should prevent git from converting line-endings.
+# (On Windows, git is often configured to automatically convert between the two - See https://git-scm.com/docs/git-config#git-config-coreautocrlf )
+
+*.js text eol=lf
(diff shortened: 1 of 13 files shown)


nfriedly:
Hi Rémy, thanks! 

I'll do a more thorough review, but at first glance this looks really great!

Please pull in the latest changes form master, though. There was a typo in master when you forked that's causing the tests to fail, but that should be fixed now. (I think GitHub even provides a button to do it for you on the page here).

Themandunord:
Hum, I don't see this button :/

Themandunord:
Ok, I do that manually :)

Themandunord:
I add some APIs : 

Get user input examples
Create user input example
Delete user input example
Get user input example
Update user input example

Themandunord:
@nfriedly I make your changes :+1: 

But I broke some tests with "trickey" text. It's a good example :+1: 
But I don't suceed to avoid the error. Have you got idea ?

nfriedly:
So, we did not need to add encodeUriComponent, the requestWrapper automatically does that when parsing the path. It turns out that the only character causing an issue was the `\` - so I removed that from the test cases and I'm going to reach out to the service team. 

This should be released in v2.23.0 once the CI loop finishes!

Thanks again for the great work!

PR #413, opened 2017-03-07
Add possibility to set another user agent for the API requests

yrezgui:
This allows to set a custom user agent for the API requests made to IBM Watson services.

##### Checklist

### DIFF
--- a/lib/requestwrapper.js
+++ b/lib/requestwrapper.js
@@ -153,8 +153,9 @@ function createRequest(parameters, _callback) {
   // Headers
   options.headers = extend({}, options.headers);
 
-  if (!isBrowser && !options.headers['User-Agent']) {
-    options.headers['User-Agent'] = pkg.name + '-nodejs-' + pkg.version;
+  if (!isBrowser) {
+    var suffix = options.headers['User-Agent'] ? ' ' + options.headers['User-Agent'] : '';
+    options.headers['User-Agent'] = pkg.name + '-nodejs-' + pkg.version + suffix;
   }
 
   // Query params
--- a/lib/requestwrapper.js
+++ b/lib/requestwrapper.js
@@ -153,7 +153,7 @@ function createRequest(parameters, _callback) {
   // Headers
   options.headers = extend({}, options.headers);
 
-  if (!isBrowser) {
+  if (!isBrowser && !options.headers['User-Agent']) {
     options.headers['User-Agent'] = pkg.name + '-nodejs-' + pkg.version;
   }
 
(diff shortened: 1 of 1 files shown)


lib/requestwrapper.js
```diff
@@ -153,7 +153,7 @@ function createRequest(parameters, _callback) {
   // Headers
   options.headers = extend({}, options.headers);
 
-  if (!isBrowser) {
+  if (!isBrowser && !options.headers['User-Agent']) {
```
germanattanasio: We should add the `User-Agent` that the user sends us but not replace the one we are using.

lib/requestwrapper.js
```diff
@@ -154,7 +154,8 @@ function createRequest(parameters, _callback) {
   options.headers = extend({}, options.headers);
 
   if (!isBrowser) {
-    options.headers['User-Agent'] = pkg.name + '-nodejs-' + pkg.version;
+    var suffix = options.headers['User-Agent'] ? ' ' + options.headers['User-Agent'] : '';
+    options.headers['User-Agent'] = pkg.name + '-nodejs-' + pkg.version + suffix;
```
germanattanasio:
```js
var providedAgent = options.headers['User-Agent']  ? ( '; ' + options.headers['User-Agent'] ) : '';
options.headers['User-Agent'] = pkg.name + '-nodejs-' + pkg.version + providedAgent;
```

lib/requestwrapper.js
```diff
@@ -153,8 +153,9 @@ function createRequest(parameters, _callback) {
   // Headers
   options.headers = extend({}, options.headers);
 
-  if (!isBrowser && !options.headers['User-Agent']) {
-    options.headers['User-Agent'] = pkg.name + '-nodejs-' + pkg.version;
+  if (!isBrowser) {
+    var suffix = options.headers['User-Agent'] ? ' ' + options.headers['User-Agent'] : '';
```
germanattanasio: you need a `;` to separate the different values

yrezgui:
@germanattanasio Done 👍

nfriedly:
I'm iffy about this for a few reasons.

First, @yrezgui: why do you want change the User-Agent header?

Second, it already works if you use the "normalized" (lowercased) version (`options.headers['user-agent']`).

Finally @germanattanasio & @yrezgui: I'm not sure about append vs overwrite, but if we're going with append, then it should have a different parameter name, maybe `userAgentSuffix` or something.

nfriedly:
I'm going to close this for now, we can come back to it later if we need to.

PR #432, opened 2017-04-17
Add new audio format to speech-to-text

mamoonraja:
Add support and unit tests for webm format to speech-to-text


test/unit/test.speech_to_text.v1.js
```diff
@@ -223,6 +223,27 @@ describe('speech_to_text', function() {
     });
   });
 
+  describe('recognizeWebM()', function() {
+    const session_path = '/v1/sessions/foo/recognize';
+    const payload = {
```
nfriedly: Since the `session_path` and `payload` are only getting used in one test, I'd recommend moving them into that test.

test/unit/test.speech_to_text.v1.js
```diff
@@ -223,6 +223,27 @@ describe('speech_to_text', function() {
...
+      audio: fs.createReadStream(__dirname + '/../resources/sample1.webm'),
+      content_type: 'audio/webm'
+    };
+
+    it('should generate a valid payload with session', function() {
+      const req = speech_to_text.recognize(extend({ session_id: 'foo' }, payload), noop);
+      assert.equal(req.uri.href, service.url + session_path);
+      assert.equal(req.method, 'POST');
+      assert.equal(req.headers['Content-Type'], payload.content_type);
```
nfriedly: This is good, but you should add another test that doesn't set the content_type, and then asserts that it is correctly detected as `audio/webm`.

test/unit/test.speech_to_text.v1.js
```diff
@@ -223,6 +223,27 @@ describe('speech_to_text', function() {
     });
   });
 
+  describe('recognizeWebM()', function() {
+    const session_path = '/v1/sessions/foo/recognize';
+    const payload = {
+      audio: fs.createReadStream(__dirname + '/../resources/sample1.webm'),
```
nfriedly: it may be worth creating a new variable here to hold the contents of this file instead of re-reading it for every test, though.

test/unit/test.speech_to_text.v1.js
```diff
@@ -223,6 +223,27 @@ describe('speech_to_text', function() {
...
+      const req = speech_to_text.recognize(extend({ session_id: 'foo' }, payload), noop);
+      assert.equal(req.uri.href, service.url + session_path);
+      assert.equal(req.method, 'POST');
+      assert.equal(req.headers['Content-Type'], payload.content_type);
+      assert.equal(req.src.path, payload.audio.path);
+    });
+
+    it('should check first four hexs for webm file', function() {
+      const file = fs.readFileSync(__dirname + '/../resources/sample1.webm');
+      assert.equal(file.slice(0, 4).toString(), '\u001aEߣ');
```
nfriedly: This test, while not being a bad idea, doesn't actually exercise our code - it's just checking your resource file. I'd say keep it, but change the description to something like "the sample webm file should have the expected header"

And, then, like I was saying above, there should be another test that does exercise our content-type detection code. Perhaps two, one that goes through the whole process of creating a request, and one that just calls `RecognizeStream.getContentType(fs.readFileSync(__dirname + '/../resources/sample1.webm'));`

nfriedly:
Great work, thanks!

PR #435, opened 2017-04-21
[Conversation] Update API

Themandunord:
Hi,

I update the conversation API with : 
 - Entities
 - Values
 - Synonyms
 - Logs

Rémy.

### DIFF
--- a/speech-to-text/recognize_stream.js
+++ b/speech-to-text/recognize_stream.js
@@ -217,7 +217,7 @@ RecognizeStream.prototype.initialize = function() {
   this.initialized = true;
 };
 
-RecognizeStream.prototype._read = function(/* size*/) {
+RecognizeStream.prototype._read = function() /* size*/ {
   // there's no easy way to control reads from the underlying library
   // so, the best we can do here is a no-op
 };
--- a/speech-to-text/recognize_stream.js
+++ b/speech-to-text/recognize_stream.js
@@ -217,7 +217,7 @@ RecognizeStream.prototype.initialize = function() {
   this.initialized = true;
 };
 
-RecognizeStream.prototype._read = function() /* size*/ {
+RecognizeStream.prototype._read = function(/* size*/) {
   // there's no easy way to control reads from the underlying library
   // so, the best we can do here is a no-op
 };
--- a/test/integration/test.conversation.js
+++ b/test/integration/test.conversation.js
@@ -909,6 +909,24 @@ describe('conversation_integration', function() {
     });
   });
 
+  describe('getLogs()', function() {
+    it('should return logs', function(done) {
+      const params = {
+        workspace_id: workspace1.workspace_id,
+        export: true,
+        page_limit: 1
+      };
+
+      conversation.getLogs(params, function(err, result) {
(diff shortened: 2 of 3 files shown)


speech-to-text/recognize_stream.js
```diff
@@ -217,7 +217,7 @@ RecognizeStream.prototype.initialize = function() {
   this.initialized = true;
 };
 
-RecognizeStream.prototype._read = function(/* size*/) {
+RecognizeStream.prototype._read = function() /* size*/ {
```
nfriedly: This change should be reverted (if the linter did it, then you may have an outdated version of `pretier` installed - run `npm update` and then run `npm run autofix`)

test/integration/test.conversation.js
```diff
@@ -594,6 +624,360 @@ describe('conversation_integration', function() {
...
+          return done(err);
+        }
+        assert.equal(result.value, test_value_update.value);
+        done();
+      });
+    });
+  });
+
+  describe('createSynonym()', function() {
+    it('should create a synonym', function(done) {
```
nfriedly: When I run it locally, this test fails with

```json
{
  "error": {
    "error": {
      "text": [
        "Text can not be empty."
      ]
    }
  }
}
```

(Actually, the error it throws is `[object Object]` but I'll fix that part in master)

And, then this failure causes a cascade where nearly all of the other synonym tests fail.

nfriedly: Op, this is because my auth.js has an older version_date (2016-09-20) - the current one (2017-04-21) is required to use this feature, although just flipping the switch seems to cause some errors elsewhere. I'll dig into it a bit on my side, though.

Themandunord:
Hi @nfriedly,

Yeah you right for the `npm update` 👍 

And yes, my auth.js is : 

```
module.exports = {
  conversation: {
    url: 'https://gateway.watsonplatform.net/conversation/api',
    username: 'XXXXX',
    password: 'XXXXX',
    workspace_id: 'XXXXX',
    version_date: '2017-04-21',
    headers: {
      'X-Watson-Learning-Opt-Out': 1,
      'X-Watson-Test': 1
    },
    version: 'v1'
  },
  // other services would go here
};
```

So on local, I have : 
image

So I don't know :/

test/integration/test.conversation.js
nfriedly: Nevermind, I had just copy-pasted the new version date incorrectly. I corrected it and now everything is passing. Great work!

Themandunord:
Ok 👍 

And how to handle the security check ?

nfriedly:
It's saying there's a potential security error in an older version of `request` - but I've already updated the SDK to the current version. I think it may be because of some lib that we depend on that in turn depends on an older version of `request`. I'm going to look into that further outside of this PR. No changes are needed from you though.

nfriedly:
This should be released shortly in v2.30.0

nfriedly:
I found the dependency responsible for the security failure and sent a PR: https://github.com/lbdremy/solr-node-client/pull/200

PR #437, opened 2017-04-21
Visual Recognition uploads a file of JSON param upgrades

jonpspri:
The 'visual_recognition.classify' service did not recognize the 'classifer_ids' param attribute and instead invoked the API with the 'default' classifier_id when POSTing a file to the API service.  I traced the but to a multi-part post not properly processing a string in the formData sent via the Request.  Replacing the String with a Buffer allowed the request to process properly.


jonpspri:
Closing this request; bug reappeared and am still investigating.

PR #438, opened 2017-04-21
Visual Recognition Classify Uploaded File

jonpspri:
It appears that order matters when Visual Recognition is comparing an uploaded file, the 'parameters' section of the multipart form must be POSTed before the file is posted.  Otherwise the classification will succeed without error (maybe test succeeds), but will be against the 'default' training, not the requested classification scheme.

The change inverts the order of the 'parameters' and 'image_files' attributes of the formData for this API call.  It's not pretty and the spec doesn't guarantee order of key extracts, but it seems to cooperate on my test setup.


nfriedly:
Hi @jonpspri 

Thanks for sending this and sorry it took me a few days to look at it. This one is strange, and I'm poking the VR team for more info, but I'm happy to merge it in.

Do you think you could help me set up an integration test that fails without this change but succeeds with it? It's a bit of a pain to run integration tests right now (I have a plan to make that better but no time...) but I can walk you through it:

1. Install mocha globally to make things easier `npm install --global mocha`

2. Create a file at `test/resources/auth.js` with the following contents:
```js
module.exports = {
  visual_recognition: {
    v3: {
      url: 'https://gateway-a.watsonplatform.net/visual-recognition/api',
      api_key: '<API KEY>',
      version_date: '2016-05-19',
      headers: {
        'X-Watson-Learning-Opt-Out': 1,
        'X-Watson-Test': 1
      },
      version: 'v3'
    }
  },
}
```

You can use your own API key, or, if you'd prefer, email me (nfriedly@us.ibm.com) and I'll create a temporary one for you. This file is .gitignore'd

3. Write your test

4. Run the command `mocha tests/integration/test.visual_recognition.js --grep "your test name"` to execute just your test

Just be aware that `npm test` will now fail because the presence of an `auth.js` triggers *all* of the integration tests.  `npm run test-unit` will run linting and the non-integration tests. 

Also, run `npm run autofix` once to clean up lint stuff (formatting, at least, can be corrected automatically)

jonpspri:
Sorry it took me so long to reply to your response.  Agreed this is a weird one and appreciate your poking at the VR team.  I'll work on the integration test when I get out of my current conference -- will let you know when it's ready to go.

jonpspri:
Okay, mostly complete.  This has one of the more aesthetically pleasing diff pages I've ever seen.

The original problem disappeared, but I ended up creating a new integration test component for the VR called 'test.visual_recognition.prepop' (for prepopulated).  I also assumed a new stanza in the auth.js file called visual_recognition_prepop, because I imagined this set of tests may want to use a different VR Service instance than the core visual_recognition test.

The test is based on the idea of testing classification against a pre-populated classifier, but it will take the time to make the classifier and wait for it to train if the classifier does not already exist.  This is not an accuracy test -- it tests that the classification results are semantically valid; not that they make sense.

Not sure what else you need to know.  Reply here and/or Slack me for additional updates.

test/integration/test.visual_recognition.prepop.js
```diff
@@ -0,0 +1,141 @@
...
+    });
+  };
+
+  before(function() {
+    //  NOTE: Use a separate auth configuration item to prevent overwrite
+    //        in the 'test.visual_recognition.js' set of tests.
+    visual_recognition = watson.visual_recognition(auth.visual_recognition_prepop.v3);
+    nock.enableNetConnect();
+
+    //  WOW.  I never thought I'd have to learn Promises just to write a test prep.  -JPS
```
nfriedly: :P

jonpspri:
Made sure all images included are public domain.  Added a shell script to re-create the zip files from the URL's in associated source txt files in case we need to swich it up.  Tinkered more with the test scripts based on lessons learned.  Shout-out to the Watson VR guys who gave me a bit more capacity in my VR service.

test/integration/test.visual_recognition.prepop.js
```diff
@@ -0,0 +1,163 @@
...
+  this.slow(TWO_SECONDS);
+  this.retries(5);
+
+  let visual_recognition;
+  let classifier_id;
+
+  const test_training_status = function(resolve, reject) {
+    //  This evil recursive function will be used to verify that the classifier
+    //  has finished training. 'resolve' and 'reject' are functions from an
+    //  enclosing promise (or a follow-on callback for resolve if you prefer)
```
nfriedly: I may end up moving this to the lib itself, because I can imagine end-users finding it useful also.

test/resources/visual_recognition.prepop/source_pictures.sh
```diff
@@ -0,0 +1,22 @@
...
+#  Creative Commons licensed files acceptable for commercial reuse by Google
+#  image search.
+
+for i in *.txt; do
+  class=$(basename "$i" .txt)
+  echo "Processing ${class}..."
+  cat ${class}.txt | while read url; do
+    test -d images/$class || mkdir -p images/$class
+    (cd images/$class || exit
+    curl -O $url)
```
nfriedly: Nice! Ability to recreate the .zips without including all of the image fines. I like it!

PR #446, opened 2017-04-27
[discovery] Add sort parameter for Discovery query

maniax89:
### Description

Adds the `sort` parameter to the Discovery `query` method -> see https://watson-api-explorer.mybluemix.net/apis/discovery-v1#!/Queries/get_v1_environments_environment_id_collections_collection_id_query for details

##### Checklist
- [X] `npm test` passes (tip: `npm run autofix` can correct most style issues)
- [X] tests are included
- [X] documentation is changed or added
- [X] link to public docs when adding new a service or new features for an existing service

##### New version_date Checklist
- [X] A new constant is avaliable with the version_date
- [X] The new constant has a comment that summarizes the changes and/or links to relevant doc pages
- [X] Any older version_date constants remain intact
- [x] The error message thrown if the service is created without a version_date indicates the new version_date constant
- [X] The example in the README includes the new version_date constant
- [X] Any relevant code in the examples/ folder has been updated to use the new version_date constant
- [X] Most tests are updated to the new version_date
- [x] 1-2 new tests are added that use the old version_date (optional, but preferred) - (no behavior is changed before/after the version date change to show any sort of different results)

### DIFF
--- a/discovery/v1.js
+++ b/discovery/v1.js
@@ -31,7 +31,7 @@ function DiscoveryV1(options) {
 
   // Check if 'version_date' was provided
   if (typeof this._options.version_date === 'undefined') {
-    throw new Error('Argument error: version_date was not specified, use DiscoveryV1.VERSION_DATE_2016_12_15');
+    throw new Error('Argument error: version_date was not specified, use DiscoveryV1.VERSION_DATE_2017_04_27');
   }
   this._options.qs.version = options.version_date;
 }
@@ -46,6 +46,11 @@ DiscoveryV1.URL = 'https://gateway.watsonplatform.net/discovery/api';
  * @type {string}
  */
 DiscoveryV1.VERSION_DATE_2016_12_15 = '2016-12-15';
+/**
+ * Release exposing the `sort` parameter on the `/query` endpoint
+ * @type {string}
+ */
+DiscoveryV1.VERSION_DATE_2017_04_27 = '2017-04-27';
 
 /**
  * Return the list of environments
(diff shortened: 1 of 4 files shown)


discovery/v1.js
```diff
@@ -445,12 +445,13 @@ DiscoveryV1.prototype.deleteDocument = function(params, callback) {
...
  * @param {String} [params.aggregation] An aggregation search uses combinations of filters and query search to return an exact answer. Aggregations are useful for building applications, because you can use them to build lists, tables, and time series. For a full list of possible aggregrations, see the Query reference.
  * @param {Number} [params.count=10] Number of documents to return
- * @param {String} [params.return] A comma separated list of the portion of the document hierarchy to return.
  * @param {Number} [params.offset=0] For pagination purposes. Returns additional pages of results. Deep pagination is highly unperformant, and should be avoided.
+ * @param {String} [params.return] A comma separated list of the portion of the document hierarchy to return.
+ * @param {String} [params.sort] A comma separated list of fields in the document to sort on. You can optionally specify a sort direction by prefixing the field with - for descending or + for ascending. Ascending is the default sort direction if no prefix is specified.
```
kognate: Is sorting supported in the newest version and should the default version string change?

maniax89: so its available in the newest version, just not sure what to do about the version string. i.e. it will work with any version date string since it isn't changing an existing parameter behavior, but adding a new one

kognate:
So if I pass the old version string it will still sort?

maniax89:
correct

maniax89:
At some point, `version=2017-04-27` will do something on the query side when we do an api-breaking change. for now, it doesn't break existing behavior therefore the version date doesn't have any affect on the API call for this `sort=` parameter. But I can bump the default version to be consistent for how we would like it to work

PR #466, opened 2017-06-01
Add content language parameter to Tone Analyzer

shubhanjan99:
Add content language as a parameter to Tone Analyzer

##### Checklist

- [x] `npm test` passes
- [x] tests are included

### DIFF
--- a/test/unit/test.tone_analyzer.v3.js
+++ b/test/unit/test.tone_analyzer.v3.js
@@ -68,7 +68,8 @@ describe('tone_analyzer.v3', function() {
     const options = {
       text: tone_request.text,
       tones: 'emotion',
-      sentences: true
+      sentences: true,
+      language: 'en'
     };
     const req = tone_analyzer.tone(options, noop);
     const body = Buffer.from(req.body).toString('ascii');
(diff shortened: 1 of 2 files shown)


tone-analyzer/v3.js
```diff
@@ -70,6 +71,7 @@ ToneAnalyzerV3.prototype.tone = function(params, callback) {
     return;
   }
   const contentType = params.isHTML ? 'text/html' : 'text/plain';
+  const contentLanguage = params.language ? params.language : 'en';
```
nfriedly: No need to set a default here. If there is no params.language, then the header should be omitted.

test/unit/test.tone_analyzer.v3.js
```diff
@@ -68,7 +68,8 @@ describe('tone_analyzer.v3', function() {
     const options = {
       text: tone_request.text,
       tones: 'emotion',
-      sentences: true
+      sentences: true,
+      language: 'en'
```
nfriedly: Please make this a new test instead of modifying the existing one. Also, please add an assert to the new test to validate that the header was actually set.

Issue #443, opened 2017-04-26
v3 Goals


nfriedly:
Breaking changes:
- [x] Fix default URL for Language Translator #318
- [x] Change createRecognizeStream to match service defaults - #257 
- [x] Share as much code as possible between [watson-speech](https://github.com/watson-developer-cloud/speech-javascript-sdk) and this repo
  * Or, just combine them together into one
- [x] ~Make nearly all methods return a Promise instead of a stream~
  * For methods where returning a Stream makes sense, add the word "stream" to the name
  * For methods where either one may make sense, just have two methods
- [x] ~Consider switching to `fetch` to allow simpler code and smaller browser builds~
  * Also consider restructuring request template as part of this - could allow "generic" things to be 
added without having to send them to the constructor.
  * Consider using https://github.com/mikeal/r2 or https://www.npmjs.com/package/axios under-the-hood
- [x] Consider switching to ES6 modules
- [x] Consider automated code generation from Swagger JSON for majority of SDK. See
  * https://github.com/swagger-api/swagger-js
  * https://github.com/wcandillon/swagger-js-codegen
- [x] Consider changing "get multiple things" api methods to `listThings()` instead of `getThings()` to reduce confusion with `getThing()` methods for getting a single "thing". 
  * Will want to create aliases for backwards-compatibility

Non-breaking changes that may still make sense to include in v3
- [x] Move request logic and error formatting into a series of methods on the base class
  * Should allow for generally simpler code because "weird" things can be scoped to only their module - e.g. #478 
- [x] ES5 builds for compatibility with older systems - #423 
- [x] Make better use of various ES6 features (class, template strings, etc)
- [x] Pre-build browser bundles
  * Modern bundle that retains various ES6 features such as Class syntax and native Promises
  * Compatibility bundle that uses ES5 syntax and includes pollyfills
  * Both could skip modules that are known to not work from a browser context

Most of these should probably get their own ticket, but we can do that as we go. 

This is a pretty tall order, and v3 will likely ship without including everything here, especially the "Consider..." ones.

PR #483, opened 2017-07-24
Generated Tone Analyzer service [do-not-merge]

glennrfisher:
This pull request would replace the handwritten Tone Analyzer service with a generated version of the service. The code was generated using the [watson-swagger-codegen][1] project with a [node generator][2] and [node template][3].

This code **should not be merged** because it is not backwards compatible. Instead, I am hoping to elicit feedback for the generation process. What changes should be made to the generator and its service template? If you were to review this pull request, what changes would you request before accepting it?

There are a few general decisions we made that may be worth discussing:

1. Explicit Content Type: The handwritten service currently accepts an `isHTML` parameter to select a content-type of `text/html` or `text/plain`. This is simpler for users than an explicit `content_type` parameter, but may not be compatible with future changes to the service. For example, if the service adds support for an additional content-type, then a _breaking change_ would be required to support it. For all languages that we've generated so far, we have required an explicit `content-type` parameter for any operation with more than one content-type.

2. Case Conventions: The generator will apply any case conventions to all services. It's currently configured to generate function names using `camelCase` and parameter names using `snake_case`. Is this reasonable, or should different conventions be used?

3. Function Documentation: The generator uses the types and descriptions from the Swagger specification to generate documentation. Does the generated documentation look reasonable? (Note that it's currently using model types, e.g. `ToneInput`, to support TypeScript in the near future.)

[1]: https://github.ibm.com/MIL/watson-swagger-codegen
[2]: https://github.ibm.com/MIL/watson-swagger-codegen/blob/node/src/main/java/io/swagger/codegen/WatsonNodeCodegen.java
[3]: https://github.ibm.com/MIL/watson-swagger-codegen/blob/node/src/main/resources/watson-node/service.mustache

### DIFF
--- a/test/integration/test.tone_analyzer.js
+++ b/test/integration/test.tone_analyzer.js
@@ -28,10 +28,10 @@ describe('tone_analyzer_integration', function() {
 
   it('tone()', function(done) {
     const mobydick = fs.readFileSync(path.join(__dirname, '../resources/mobydick.txt'), 'utf8');
-    tone_analyzer.tone({ text: mobydick }, done);
+    tone_analyzer.tone({ content_type: 'text/plain', tone_input: mobydick }, done);
   });
 
-  it('tone_chat()', function(done) {
+  it('toneChat()', function(done) {
     const utterances = {
       utterances: [
         { text: 'My charger isn’t working.', user: 'customer' },
@@ -43,6 +43,6 @@ describe('tone_analyzer_integration', function() {
         { text: 'I’m sorry you’re having issues with charging. What kind of charger do you have?', user: 'agent' }
       ]
     };
-    tone_analyzer.tone_chat(utterances, done);
+    tone_analyzer.toneChat(utterances, done);
   });
 });
--- a/test/unit/test.tone_analyzer.v3.js
+++ b/test/unit/test.tone_analyzer.v3.js
@@ -8,9 +8,7 @@ const extend = require('extend');
 describe('tone_analyzer.v3', function() {
   const noop = function() {};
 
-  const tone_request = {
-    text: 'IBM Watson Developer Cloud'
-  };
+  const text = 'IBM Watson Developer Cloud';
   const tone_response = {
     tree: {}
   };
@@ -33,7 +31,7 @@ describe('tone_analyzer.v3', function() {
 
   before(function() {
     nock.disableNetConnect();
-    nock(service.url).persist().post(tone_path + '?version=2016-05-19', tone_request.text).reply(200, tone_response);
+    nock(service.url).persist().post(tone_path + '?version=2016-05-19', text).reply(200, tone_response);
   });
 
   after(function() {
@@ -55,10 +53,11 @@ describe('tone_analyzer.v3', function() {
   });
 
   it('tone API should generate a valid payload with text', function() {
-    const req = tone_analyzer.tone(tone_request, noop);
+    const options = { content_type: 'text/plain', tone_input: text };
+    const req = tone_analyzer.tone(options, noop);
     const body = Buffer.from(req.body).toString('ascii');
     assert.equal(req.uri.href, service.url + tone_path + '?version=2016-05-19');
-    assert.equal(body, tone_request.text);
+    assert.equal(body, text);
     assert.equal(req.method, 'POST');
     assert.equal(req.headers['content-type'], 'text/plain');
     assert.equal(req.headers['accept'], 'application/json');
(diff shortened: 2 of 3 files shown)


nfriedly:
Hey, I like this, and I think it's definitely the direction we need to head in.  New services should get this treatment from the get-go.

However, for existing services, I think we're going to have to put a lot of work into making things backwards-compatible - all the old param names should continue to work, and all of the existing tests should pass without modification.

Can we do something with 80% generated and 20% custom code? (Preferably in the same file, but it could be broken up into generated "core" files and hand-written "compat" files I suppose.)

I'm OK if the docs and examples change, and/or we mark the old names as deprecated, but I really don't want to release something that will break a ton of users, even with a major version tag. 

It should probably coincide with a major version bump anyways, though - there's a number of changes I want to make that this would make easier. The big one is returning promises for most methods - see #443.

Also, do you have any thoughts for how to handle the couple of methods that really should return Streams - mainly Speech to Text?  I suppose if we're going with the core/compat method, the core version could just return a promise, while the compat version returned a stream.

glennrfisher:
Hey Nathan, thanks for the comments! I've included some responses below.

> I think we're going to have to put a lot of work into making things backwards-compatible...Can we do something with 80% generated and 20% custom code?

Yes! I completely agree. Sorry if I didn't communicate it well, but I didn't intend to replace the existing Conversation service with generated code. Instead, I was interested to hear your critique of the code since future services with the same generator / template would look similar.

We've been (mostly) avoiding breaking changes to the other languages. But the generator is still useful for new functionality.

For example, the Conversation service added several operations for dialog nodes. We can generate that code, then manually transfer _those particular methods_ into the handwritten SDK. That's what we did with [this pull request][1] to the Python SDK. It added the `tone_chat` function but kept the handwritten `tone` function.

If the generated code looks good, then we can use it to help you add support for service changes and new functionality.

> I really don't want to release something that will break a ton of users, even with a major version tag.

Yes, agreed. We've been holding off on updates to any single service that would cause a major version bump.

At some point, though, we suspect that the generator will be able to generate all of the REST operations. When we reach that stage, we're hoping to publish a single major version bump to transition to generated code—at least for the other SDKs. As you mentioned, that would best coincide with other breaking changes, like promises.

> It should probably coincide with a major version bump anyways, though - there's a number of changes I want to make that this would make easier. The big one is returning promises for most methods - see #443.

This is actually a great use-case for the generator! We can modify a single template, then _generate_ those changes for the entire SDK.

For example, we've already poked into TypeScript support. It would be laborious to manually add type definitions for every service. But we already added a template to generate TypeScript type definitions, which can be applied to lots of services.

> Also, do you have any thoughts for how to handle the couple of methods that really should return Streams - mainly Speech to Text? I suppose if we're going with the core/compat method, the core version could just return a promise, while the compat version returned a stream.

Yes, that's a terrific question. We haven't tackled the Speech to Text or Text to Speech services yet. We should be able to generate all of the REST endpoints without issue, but the socket API cannot be documented through Swagger.

This is a problem that we'll be running into soon. I suspect we'll keep the handwritten socket code, but find a convenient way to access it. In Swift, for example, we'll probably handwrite the socket code in a separate file, as an extension to the generated `SpeechToText` class.

[1]: https://github.com/watson-developer-cloud/python-sdk/pull/198

glennrfisher:
Just to give an example, I spent a couple minutes updating the generation template for promises:

1. Remove `callback` documentation.
2. Remove `callback` parameter.
3. Return `Promise.reject(...)` when there are missing parameters.
4. Remove `callback` parameter from `requestFactory` calls.
(Assume that `requestFactory` returns a promise.)

These are admittedly simple changes, so they wouldn't take long by hand anyway. But the generator makes it trivial. And other sweeping changes would be easy to apply if any further changes were required—for example, to document all of the functions as returning a promise, or to switch to async/await.

Here's a gist with the example: https://gist.github.com/glennrfisher/7ddbc412084dcf19413321dd0602404b

germanattanasio:
@nfriedly correct me if I'm wrong but we want to support callbacks and promises at the same time.

When calling a function if there is no callback then we return a Promise otherwise we assume the second parameter is a callback.

Supporting only callbacks is not a bad idea. Right now developers can `promisify` the object and there have a promise for each method.

nfriedly:
Cool @glennrfisher, I think we're mostly on the same page!

> We can generate that code, then manually transfer those particular methods into the handwritten SDK.

I do want to avoid that. I'd rather see a system where an update to the generated part would not involve any copy-pasting. I'm not sure exactly how it should work, but updating code for new swagger definitions should be a single command, and it would somehow take into account the hand-written code and adjust the output to include that. Can the part of the existing codebase be used in the generation? Perhaps even just a list of what files exist?

@germanattanasio 

> correct me if I'm wrong but we want to support callbacks and promises at the same time.

Yes, that's correct - that's one of those backwards-compatibility things ;)

> When calling a function if there is no callback then we return a Promise otherwise we assume the second parameter is a callback.

I was thinking we would *always* return a promise, and then if there was a callback passed in, we'd also deal with that.

> Right now developers can `promisify` the object and there have a promise for each method.

True, but I don't see it happen very much in practice. Also, I want promises for our tests, and *I* don't want to have to `promisify` it. I'd rather support both promises and callbacks.

germanattanasio:
@nfriedly One of the problems we want to fix with automated SDKs is naming. Some of our methods to list resources start with `listXXX()` and some others with `getXXX()`. 

@glennrfisher has been updating the swagger to fix problems as the one mentioned avoid, this makes the generated code consistent across languages. 

We will do some manual work to avoid breaking changes, but the end goal is to push users to the newer generated version.

We can talk with @kemerson16, and you can spend some time with Gleen on this so that the changes you have in mind are implemented in the generator.
What do you think?

nfriedly:
@germanattanasio yea, I agree. This is what I wrote on #443: 

> * [ ] Consider changing "get multiple things" api methods to listThings() instead of getThings() to reduce confusion with getThing() methods for getting a single "thing".
>   * Will want to create aliases for backwards-compatibility

I think the generator should try to always do "list", and the hand-written files should provide the "get" variation. 

This may be difficult to do with the current swagger, as I don't believe it always follows this convention. We can talk to the docs team about that.

glennrfisher:
@nfriedly Yeah! For sure. We've been modifying the `operationId` in the Swagger specification to use `list`. So all of the services that we've looked at so far (everything except the speech services now) should have been updated to use `list` instead of `get` when querying for a list of values.

nfriedly:
hey @glennrfisher - I just created a `generated` branch, can you update this PR to target that branch?

glennrfisher:
I'm going to close this issue for now.

We've made a lot of progress generating different services. Each service tends to have its own gotchas, so it's helpful to try generating them all. But it can be overwhelming to review.

Instead of creating a pull request each time we generate a new service, let's focus our review on a particular service. German suggested Natural Language Classifier, so that pull request should be ready soon.

I expect us to continue generating other services, while waiting and acting on feedback about Natural Language Classifier. We'll publish those changes in separate branches, like `typescript/conversation`, but we will wait to open a pull request.

PR #484, opened 2017-07-24
Generated Conversation service [do-not-merge]

glennrfisher:
This pull request would replace the handwritten Conversation service with a generated version of the service. The code was generated using the [watson-swagger-codegen][1] project with a [node generator][2] and [node template][3].

This code **should not be merged** because it is not backwards compatible. Instead, I am hoping to elicit feedback for the generation process. What changes should be made to the generator and its service template? If you were to review this pull request, what changes would you request before accepting it?

There are a few general decisions we made that may be worth discussing:

1. Renaming Conflicting Parameters: Some operations (e.g. `updateIntent`) have parameters with conflicting names (e.g. `intent` in path and `intent` in body). The handwritten code prefixes the conflicting path parameter with `old_`. But for consistency with other languages, the generated code prefixes the body parameters with `new_`.

2. Renaming Parameters: Some parameters are renamed from the Swagger specification. Most renaming is from a change in case (e.g. `snake_case` to `camelCase`) or to resolve a conflict (e.g. `intent` to `new_intent`).

3. Using `getMissingParams`: The handwritten service very cleanly accepts a `requiredParams` list in the `parameters` object. Unfortunately, I had trouble getting this to work when parameters are renamed from the Swagger specification. So I generated code by executing the `getMissingParams` function directly.

4. Listing a Collection: Some collections in the handwritten code are retrieved with e.g. `getExamples`. For SDK generation, we've adopted the convention of using `listExamples` to retrieve a collection of examples and `getExample` to retrieve a single example.

5. CounterExample to Counterexample: This change comes directly from the Swagger specification, which uses `Counterexample` instead of `CounterExample`. Looks like counterexample is defined as a single word, so I suppose this is a reasonable convention.

[1]: https://github.ibm.com/MIL/watson-swagger-codegen
[2]: https://github.ibm.com/MIL/watson-swagger-codegen/blob/node/src/main/java/io/swagger/codegen/WatsonNodeCodegen.java
[3]: https://github.ibm.com/MIL/watson-swagger-codegen/blob/node/src/main/resources/watson-node/service.mustache

### DIFF
--- a/conversation/v1.js
+++ b/conversation/v1.js
@@ -205,7 +205,7 @@ ConversationV1.prototype.listCounterexamples = function(params, callback) {
  */
 ConversationV1.prototype.updateCounterexample = function(params, callback) {
   params = params || {};
-  const requiredParams = ['workspace_id', 'text', ];
+  const requiredParams = ['workspace_id', 'text', 'body'];
   const missingParams = helper.getMissingParams(params, requiredParams);
   if (missingParams) {
     callback(missingParams);
@@ -247,7 +247,7 @@ ConversationV1.prototype.updateCounterexample = function(params, callback) {
  */
 ConversationV1.prototype.createEntity = function(params, callback) {
   params = params || {};
-  const requiredParams = ['workspace_id', 'entity', ];
+  const requiredParams = ['workspace_id', 'entity'];
   const missingParams = helper.getMissingParams(params, requiredParams);
   if (missingParams) {
     callback(missingParams);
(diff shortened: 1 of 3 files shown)


glennrfisher:
I'm going to close this issue for now.

We've made a lot of progress generating different services. Each service tends to have its own gotchas, so it's helpful to try generating them all. But it can be overwhelming to review.

Instead of creating a pull request each time we generate a new service, let's focus our review on a particular service. German suggested Natural Language Classifier, so that pull request should be ready soon.

I expect us to continue generating other services, while waiting and acting on feedback about Natural Language Classifier. We'll publish those changes in separate branches, like `typescript/conversation`, but we will wait to open a pull request.

PR #501, opened 2017-09-22
add passage extraction parameters

### LINKED ISSUE #500: [discovery] passage extraction parameters not supported
Following passage extraction parameters have no effect with current version of SDK.

```
queryParams['passages.fields'] = 'description,text';
queryParams['passages.characters'] = 2000;
queryParams['passages.count'] = 15;
```

When DIscovery url is called directly using curl or similar tool with these parameters the passages are extracted correctly. Is there a timeline for adding the parameters to the SDK?

don-tjandra:
Add following parameter for passage extraction.

passages.fields: A comma-separated list of fields in the index that passages will be drawn from.
passages.count: The maximum number of passages to return. 
passages.characters: The approximate number of characters that any one passage should have.

Fixes #500

### DIFF
--- a/discovery/v1.js
+++ b/discovery/v1.js
@@ -680,7 +680,21 @@ DiscoveryV1.prototype.query = function(params, callback) {
       method: 'GET',
       json: true,
       path: pick(params, ['environment_id', 'collection_id']),
-      qs: pick(params, ['query', 'natural_language_query', 'filter', 'aggregation', 'count', 'offset', 'return', 'sort', 'passages', 'highlight', 'passages.fields', 'passages.count', 'passages.characters'])
+      qs: pick(params, [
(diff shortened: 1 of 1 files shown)


nfriedly:
Hey, thanks for this and sorry for the slow response. Can you fix the lint errors? I think just running `npm run autofix` (which just runs `eslint . --fix` under the covers) will do the trick.

don-tjandra:
Latest commit passed lint checks but hit this security blocker

https://nodesecurity.io/advisories/536

Is there something I can do on my end?

nfriedly:
Yes, the string.js security vulnerability has already been addressed in a separate PR, so this should be good to go now.

It should be released by CI shortly.

Thanks again!

PR #506, opened 2017-09-27
remove dep "string"; closes #502

### LINKED ISSUE #502: security advisory: vulnerability in dependency "string"
References:

- [Snyk](https://snyk.io/vuln/npm:string:20170907)
- [NSP](https://nodesecurity.io/advisories/536)

as of this writing, there are no fixes.

My recommendation is to drop the dependency (it's only used [here](https://github.com/watson-developer-cloud/node-sdk/blob/1557aca4a2378fab120a4684ac4b5305142155a4/lib/requestwrapper.js#L42)) and use a ES6 template string or something else.

boneskull:

This removes dependency `string` since it wasn't being used for much, and there's currently no upgrade path to avoid the vulnerability.

This code is covered by many existing tests.

### DIFF
--- a/lib/requestwrapper.js
+++ b/lib/requestwrapper.js
@@ -20,7 +20,6 @@ const extend = require('extend');
 const request = require('request');
 const pkg = require('../package.json');
 const helper = require('./helper');
-const parseString = require('string');
 const readableStream = require('stream').PassThrough;
 const isBrowser = typeof window === 'object';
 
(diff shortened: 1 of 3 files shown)


nfriedly:
Yea, we run NSP on pull requests :P 

Probably should look into running it even when the code didn't change. 

Not sure why the one test failed, looks like a storage hiccup in travis ci and nothing to do with your code.

Thank you for the fix! I'll get it out in a patch release shortly.

PR #509, opened 2017-10-03
:bug: Fixed bug with calling options.url if options param was null

anweshan:
Previously options.url was called even if options was null, causing error of options.url being undefined rather than options being undefined.

##### Checklist

- [ x] `npm test` passes (tip: `npm run autofix` can correct most style issues)
- [ ] tests are included (Previous tests already test for missing parameter options)

### DIFF
--- a/language-translator/v2.js
+++ b/language-translator/v2.js
@@ -38,7 +38,7 @@ function LanguageTranslatorV2(options) {
   // and correcting the default URL here will break older code, so it must be reserved for a major release.
   // todo: consider checking for options.url === LanguageTranslationV2.URL and also throw this warning then.
   // (This probably does't matter since the api didn't change)
-  if (!options.url) {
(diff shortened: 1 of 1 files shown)


language-translator/v2.js
```diff
@@ -38,7 +38,7 @@ function LanguageTranslatorV2(options) {
   // and correcting the default URL here will break older code, so it must be reserved for a major release.
   // todo: consider checking for options.url === LanguageTranslationV2.URL and also throw this warning then.
   // (This probably does't matter since the api didn't change)
-  if (!options.url) {
+  if (options && !options.url) {
```
nfriedly: Make this `if (!options || !options.url) {` - we want to give the warning if the options object is omitted entirely.

anweshan: Ok, agreed, just made this change

Issue #324, opened 2016-10-16
[text-to-speech] Add "Repair WAV Header" (as in the ios-sdk)


tanmayb123:
In the ios-sdk on the WDC, for the TtS service, there's functionality that can repair the WAV Header once it's received from Watson. I'd like to add that to the Node SDK.

Steps to reproduce: 
1 - Use the following code:

``` javascript
     var text_to_speech = new TextToSpeechV1({
        username: '[YOUR_USERNAME]',
        password: '[YOUR_PASSWORD]'
      });
      var params = {
        text: response.output.text[0],
        voice: 'en-US_MichaelVoice', // Optional voice
        accept: 'audio/wav'
      };
      var stream = text_to_speech.synthesize(params).pipe(fs.createWriteStream('output.wav'));
      stream.on('finish', function() {
        console.log('finished!');
      })
```

2 - Run the code, and open "output.wav" in iTunes.
3 - It plays, but if you notice, the duration of the file is ~20 hours long...
**NOTE - It does play properly, and when the file is "over" (Watson is done speaking), iTunes does stop playing. It does NOT play for ~20 hours. Although, I think this may depend on the media player you use.**

I know the cause of this problem: The header of the WAV file isn't able to indicate the "length" of the file properly. This is causing me problems when trying to play the WAV file from Node.

The ios-sdk handles this nicely, by repairing the header of the file. In fact, here's an excerpt of some code from the ios-sdk:

``` swift
func dataToUTF8String(data: NSData, offset: Int, length: Int) -> String? {
    let range = NSMakeRange(offset, length)
    let subdata = data.subdata(with: range)
    return String(data: subdata, encoding: String.Encoding.utf8)
}

func dataToUInt32(data: NSData, offset: Int) -> Int {
    var num: UInt32 = 0
    let length = 4
    let range = NSMakeRange(offset, length)
    data.getBytes(&num, range: range)
    return Int(num)
}

//This is the actual function that repairs the Header:
func repairWAVHeader(data: NSMutableData) {

    // resources for WAV header format:
    // [1] http://unusedino.de/ec64/technical/formats/wav.html
    // [2] http://soundfile.sapp.org/doc/WaveFormat/

    // update RIFF chunk size
    let fileLength = data.length
    var riffChunkSize = UInt32(fileLength - 8)
    let riffChunkSizeRange = NSMakeRange(4, 4)
    data.replaceBytes(in: riffChunkSizeRange, withBytes: &riffChunkSize)

[...]

PR #525, opened 2017-10-25
[text-to-speech] Repair WAV headers. #324

dpopp07:
To resolve Issue #324, I wrote a function to intercept the stream, calculate the proper file size, edit the stream, and return it.

As I did this, I realized that it is not really possible without returning a Promise - reading/editing the stream involves an asynchronous process that finishes _after_ the function ends and the user calls `pipe()` on the returned stream (which is undefined until the asynchronous process finishes).

Glenn mentioned that we were hoping to use Promises in the next major release of the SDK and suggested that I make a PR with my change, just commented out for now, so that we have the code when we want it.

If the user expects a Promise when using this service, this function accurately repairs the WAV header.
I did look at the `wav` package after reading the comments attached to this issue. The package does not quite do what we need for this issue. It actually expects that there are only 2 subchunks in the WAV file and doesn't account for metadata chunks, which this code does using the same method as the Swift SDK.

Please reply with any questions/comments/suggestions. Thanks.

##### Checklist

- [ ] tests are included

- [ ] Most tests are updated to the new version_date
- [ ] 1-2 new tests are added that use the old version_date (optional, but preferred)

### DIFF
--- a/README.md
+++ b/README.md
@@ -548,8 +548,18 @@ var params = {
   accept: 'audio/wav'
 };
 
-// Pipe the synthesized text to a file
-text_to_speech.synthesize(params).pipe(fs.createWriteStream('output.wav'));
+// Synthesize speech, correct the wav header, then save to disk
+// (wav header requires a file length, but this is unknown until after the header is already generated and sent)
+textToSpeech
+  .synthesize(params, function(err, audio) {
+    if (err) {
+      console.log(err);
+      return;
+    }
+    textToSpeech.repairWavHeader(audio);
+    fs.writeFileSync('audio.wav', audio);
+    console.log('audio.wav written with a corrected wav header');
+});
 ```
 
 ### Tone Analyzer
--- a/examples/text_to_speech.v1.js
+++ b/examples/text_to_speech.v1.js
@@ -10,7 +10,29 @@ const textToSpeech = new TextToSpeechV1({
   // password: 'INSERT YOUR PASSWORD FOR THE SERVICE HERE'
 });
 
+// Synthesize speech, correct the wav header, then save to disk
+// (wav header requires a file length, but this is unknown until after the header is already generated and sent)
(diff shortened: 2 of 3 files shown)


nfriedly:
Cool, this is great work!

Regarding the switch to promises, I think that for things like this where a stream actually makes sense, we'll have two variations:

* a promise-based one where it does similar to your code and loads the entire contents into memory, then fixes the header, then resolves with the data (no need for `into-stream` though)
* a streaming version similar to the current output, where the header is just left incorrect

...although, now that the service supports websockets, it might make more sense to only do promises for the HTTP version and streams for the websocket one. (The SDK has not yet implemented websocket support, though :(

But, never mind all of that. Right now, all SDK methods both accept a callback and return a stream, so I think it would make more sense to work with the callback variant. Then there's no need to buffer and concatenate all of the data chunks, and no need to turn it back into a stream when you're done. It might even make sense to just kick it on automatically, although I'm not sure I want the SDK to do that due to the potential for confusion.

Can you do this:
* Expose the method as `TextToSpeechV1.prototype.repairWaveHeader()` or something similar
* Drop support for streams and promises (for now); just make it a synchronous method that accepts a buffer and modifies it in place.
* Add an example that shows it in action. I'm imagining it will look something like this:

```js
// Synthesize speech, correct the wav header, then save to disk
// (wav header requires a file length, but this is unknown until after the header is already generated and sent)
textToSpeech
  .synthesize({
    text: 'Hello from IBM Watson',
    accept: 'audio/wav'
  }, function(err, audio) {
    if (err) {
      console.log(err);
      return;
    }
   textToSpeech.repairWavHeader(audio);
   fs.writeFileSync('audio.wav', audio);
   console.log('audio.wav written with a corrected wav header');
})

* And, of course, uncomment everything :)

dpopp07:
Thanks for the comments! I made this change and it works with your example.

When you say 'Add an example', do you mean in the comments? Or in the README?

PR #526, opened 2017-10-26
Added file handling for node-codegen

AmmarD212:
This is my initial attempt at handling files somewhat reasonably and uniformly for NodeJS. The idea is for parameters that are files, allow the user to pass in a ReadableStream or an Object, where the Object in question here is a `form-data` object. For details see (https://github.com/request/request#forms)

The main critical piece that needs review and feedback is the `buildRequestFileObject` function. This function builds the form-data object for each file parameter. 

For `filename`, we grab the `filename` field of the `form-data` Object if that's what the user supplied. If the user supplied a ReadableStream, then we use `path` property of the stream. If all fails, the `filename` decays to `null`.

For `contentType`, we have two cases. If there's only a single producer, then the `contentType` is set to be whatever that is unless the user supplied the file parameter as a `form-data` Object with  a `contentType` property. For multiple producers, we set the `contentType` to the user specified value in either the `contentType` field of a `form-data` Object, or the generated `<paramName>_content_type` parameter. If all fails, the `contentType` always decays to `application/octet-stream`.

### DIFF
--- a/natural-language-classifier/v1.js
+++ b/natural-language-classifier/v1.js
@@ -1,31 +1,74 @@
 /**
- * Copyright 2015 IBM Corp. All Rights Reserved.
+ * Copyright 2017 IBM All Rights Reserved.
  *
- * Licensed under the Apache License, Version 2.0 (the "License"); you may not
- * use this file except in compliance with the License. You may obtain a copy of
- * the License at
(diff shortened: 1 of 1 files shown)


germanattanasio:
This looks good and we talked about moving this to the core sdk so I'm going to close this and wait for the other PR

PR #530, opened 2017-11-03
Generate Natural Language Classifier

glennrfisher:
This pull request is for a generated version of Natural Language Classifier.

#### Key Files

- `natural-language-classifier/v1-generated.ts` - Generated from Swagger.
- `natural-language-classifier/v1.ts` - Handwritten adapter for existing API.
- `test.natural_language_classifier.v1.generated.js` - Modified tests for generated API.

#### Feedback

Here is some specific feedback that would be helpful for refining our generator.

- [ ] Does the generated API look reasonable? (e.g. `params.headers` were collapsed into `params`)
- [ ] Does the generated method implementation look reasonable? (e.g. not using `pick`)
- [ ] Does the generated documentation look reasonable? (e.g. method documentation)
- [ ] Do the generated parameter interfaces look reasonable? (e.g.` ClassifyParams`)
- [ ] Do the generated response model interfaces look reasonable? (e.g. `Classifier`)
- [ ] Does the handwritten adapter look reasonable? (i.e. `v1.ts`)


glennrfisher:
I wasn't quite sure how to implement the adapter from `create` to `createClassifier`. I left it as a `TODO`. Although it turns out that all of the tests pass.

AmmarD212:
@glennrfisher  The tests pass for createClassifier because I changed the tests when I created the `typescript/natural-language-classifier`. If you take a look at the original tests here [original](https://github.com/watson-developer-cloud/node-sdk/blob/master/test/unit/test.natural_language_classifier.v1.js), you'll see an additional test that tests passing an array of JSON as `training_data`. I stripped that test out because I was merely testing the base functionality, not backwards compatibility at that point.

AmmarD212:
Currently, the generated datatype for file parameters is `ReadableStream| Object |UInt8Array`.  The `Object` in question here refers to a formData object. Since we're going to be displaying TypeScript code in the Github repo, I thought it may make sense to write an interface that describes this object's schema and add it to the generated code. Something like the following

`
interface FileObject {
    value: ReadableStream | UInt8Array | string;
    options: FileOptions;
}

interface FileOptions {
    filename: string;
    contentType: string;
}
`

I'd love to know your guys's thoughts on this.

nfriedly:
@AmmarD212 I think that's a good idea, although I suspect it already exists somewhere.  Making our own is probably fine for now, though.

natural-language-classifier/v1-generated.ts
```diff
@@ -0,0 +1,296 @@
...
+   * List classifiers.
+   *
+   * Returns an empty array if no classifiers are available.
+   *
+   * @param {Object} [params] - The parameters to send to the service.
+   * @param {Function} [callback] - The callback that handles the response.
+   */
+  listClassifiers(params?: GeneratedNaturalLanguageClassifierV1.ListClassifiersParams, callback?: GeneratedNaturalLanguageClassifierV1.Callback<GeneratedNaturalLanguageClassifierV1.ClassifierList>): ReadableStream | void {
+    params = params || {};
+    if (typeof params === 'function' && !callback) {
```
germanattanasio: 🤔

germanattanasio: I think we always want to have parameters and callback.

germanattanasio: Remember that `list()` could have parameters in the future

natural-language-classifier/v1.ts
```diff
@@ -0,0 +1,80 @@
...
+   */
+  create(params, callback) {
+    // TODO
+  }
+
+  /**
+   * Returns the classification information for a classifier on a phrase
+   */
+  classify(params, callback) {
+    params = params || {};
```
germanattanasio: don't modify arguments

germanattanasio: you can call it `_params`

natural-language-classifier/v1.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  /**
+   * Returns the training status of the classifier
+   */
+  status(params, callback) {
+    params = params || {};
+    if (!params.classifier_id) {
+      params.classifier_id = params.classifier;
+    }
+    super.getClassifier(params, callback);
```
germanattanasio:
```js
return super.getClassifier(params, callback);
```

natural-language-classifier/v1.ts
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  /**
+   * Retrieves the list of classifiers for the user
+   */
+   list(params, callback) {
+     params = params || {};
+     super.listClassifiers(params, callback);
```
germanattanasio:
```
return super.listClassifiers(params || {}, callback);
```

natural-language-classifier/v1.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  /**
+   * Deletes a classifier
+   */
+  remove(params, callback) {
+    params = params || {};
+    if (!params.classifier_id) {
+      params.classifier_id = params.classifier;
+    }
+    super.deleteClassifier(params, callback);
```
germanattanasio:
```
return super.deleteClassifier(Object.assign({ classifier_id : (params ? params.classifier : null}, params), callback);
```

natural-language-classifier/v1.ts
glennrfisher: Doh! Good catch, thanks.

natural-language-classifier/v1.ts
glennrfisher: Doh! Good catch, thanks.

natural-language-classifier/v1.ts
glennrfisher: Doh! Good catch, thanks.

natural-language-classifier/v1-generated.ts
glennrfisher: This is definitely awkward. I really wish JavaScript supported function overloading.

It's a fair point that new parameters may be added to the `listClassifiers` function. A new _required_ parameter will necessitate a breaking change for all SDKs, regardless. However, a new _optional_ parameter would still keep the entire `params` object as optional.

Personally, I'd rather embrace the _internal_ awkwardness of checking for the `params` argument rather than the _external_ awkwardness of requiring users to pass a `params` argument of `{}` or `undefined`.

Would be helpful to get @nfriedly's opinion when he's back from his trip.

germanattanasio: For consistency, we should stick with `(params, callback)` for all our methods.
You can leave it if you want but it won't be documented.
In Java, we had a similar situation, and we decided to send a model in the request even if it was empty or null.

glennrfisher: I talked to Mike to learn more about this situation in Java. While chatting, we realized that we could create an overloaded method in Java:

```java
// original method -- options has no required parameters
public ServiceCall<Model> listModels(Options options) {
    ...
}

// new overloaded method
public ServiceCall<Model> listModels() {
    listModels(null);
}
```

This was a quick and easy change to the Java service template. Mike is going to test it out and if all goes well then he'll open a pull request in `watson-swagger-codegen`. Any generated code should be backwards-compatible with the existing Java SDK release since it doesn't modify the existing method.

In light of that change, I think it makes sense to try the equivalent of an overload in Node. I think the closest we can get is to check the arguments and swizzle them around, as awkward as it is.

natural-language-classifier/v1.ts
glennrfisher: Yup! You're right. There's even a [rule][1] for it in ESLint.

[1]: https://eslint.org/docs/rules/no-param-reassign

natural-language-classifier/v1-generated.ts
germanattanasio: Which method do we document the overloaded or the other one?

glennrfisher: I think they both have to be documented.

For example, consider a `listModels` endpoint that only has _optional_ parameters, so both methods will be generated. The `listModels()` function with no arguments may be the common path, so it should be documented. But users that want to pass specific options may choose to use `listModels(ListModelsOptions options)`. So it should be documented as well.

nfriedly: I like the code as it's written - with optional params.  

For the documentation, I think that the typescript typings and JSDoc should accurately reflect that it is optional, but any hand-written examples should probably pass in an options object for consistency.

The API ref doesn't explicitly mention the options object for Node.js, just it's keys, so it woon't really make any difference there aside from the code example.

germanattanasio: @nfriedly How do you expect us to support promises without breaking users? People in the community don't like to support callbacks and promises in the same function.

germanattanasio: Regarding the optional parameter, I didn't hear a strong argument of why we should do it so let's just not do it. Especially since we are going to document examples where we are going to pass empty objects.

glennrfisher: Good discussion, but let's leave it optional. That seems more idiomatic for Node. And for what it's worth, this is the Node equivalent of method overloading in the [Java template][1], so it encourages consistency between languages.

[1]: https://github.ibm.com/MIL/watson-swagger-codegen/blob/develop/src/main/resources/watson-java/service.mustache#L360

germanattanasio: I agree with you. Let's use your approach and make sure we don't document it in the API Reference for now

glennrfisher:
Thanks for the feedback everyone! 

I ended up creating a new branch from `typescript/master` to make sure that I had those changes and supported NLC in the same way as other services. I incorporated all of this feedback into those changes and will open a new pull request to keep things clean.

PR #534, opened 2017-11-13
Generated Discovery with Adapter

AmmarD212:
### Considerations/Notes
- `_ensureFilename` is not needed anymore with the core-sdk changes, unless we think users are directly using it. I've kept it around just in case users are actually using it directly.
- `createConfiguration` && `updateConfiguration` endpoints no longer use `multi-part` requests. They place the parameters in the `body` now.
- `createConfiguration` && `updateConfiguration` tests needed to be update such that they're passing `params.file` as JSON `Object` and not a stream/file. The sdk documents the `file` parameter as a JSON `Object` so I am assuming files were used in the tests for convenience.
- `prettier` was used to format the code. `prettier` has support for TypeScript now.


discovery/v1-generated.ts
```diff
@@ -0,0 +1,2602 @@
...
+  createEnvironment(
+    params: GeneratedDiscoveryV1.CreateEnvironmentParams,
+    callback?: GeneratedDiscoveryV1.Callback<GeneratedDiscoveryV1.Environment>
+  ): ReadableStream | void {
+    const _params = extend({}, params);
+    const _callback = typeof callback === 'function' ? callback : () => {};
+    const requiredParams = ['name'];
+    const missingParams = getMissingParams(_params, requiredParams);
+    if (missingParams) {
+      return _callback(missingParams);
```
germanattanasio: I think we can have 
```js
if (missingParams) {
      return (typeof callback === 'function') ? callback(missingParams) : null;
}
```

Instead of creating our own internal callback

discovery/v1-generated.ts
```diff
@@ -0,0 +1,2602 @@
...
+    }
+    const body = {
```
germanattanasio: why don't we use `pick` instead of constructing the object? do we have cases where body and params have different parameters names

AmmarD212: The name in params is usuall the `x-alternate-name`, while the name we're supposed to send to the service is the base name.

discovery/v1-generated.ts
AmmarD212: Glenn was suggesting this change to the template

`
{{#hasRequiredParams}}
    const _callback = (callback) ? callback : () => {};
    {{/hasRequiredParams}}
    {{^hasRequiredParams}}
    const _params = (typeof params === 'function' && !callback) ? {} : extend({}, params);
    const _callback = (typeof params === 'function' && !callback) ? params : () => {};
    {{/hasRequiredParams}}
`

AmmarD212: I know you were saying that we should do what @glennrfisher  proposed in the adapters instead of the Generated service, but we have to consider that the adapters are going to be dissolved at some point and are likely to be short-lived. Therefore, by including what @glennrfisher proposed in the generated service, we reduce the amount of work we need to do after the adapters are dissolved.

AmmarD212:
@germanattanasio I am merging this for now. Changes to the `params` and `callback` handling can be added later.

discovery/v1-generated.ts
glennrfisher: Yup! That's right. We need to construct an object to handle different naming conventions and alternate names.

PR #535, opened 2017-11-14
Generated Visual Recognition with Adapter

AmmarD212:
##  Notes/Considerations

- Tests were updated to use the new name i.e `watson_vision_combined`.
- BaseService was updated to pick up `api_keys`
- Style updates for core-sdk.


lib/base_service.ts
```diff
@@ -14,12 +14,14 @@
  * limitations under the License.
  */
 
- 'use strict';
+'use strict';
```
germanattanasio: do we need the `use strict` ?

test/unit/test.visual_recognition.v3.js
```diff
@@ -84,7 +84,7 @@ describe('visual_recognition', function() {
     nock.cleanAll();
   });
 
-  const visual_recognition = watson.visual_recognition(service);
+  const visual_recognition = watson.watson_vision_combined(service);
```
germanattanasio: why?

AmmarD212: According to swagger: `"x-vcap-service-name": "watson_vision_combined"`

lib/base_service.ts
AmmarD212: This doesn't need to be here. I can configure the `tsc` compiler to emit `use strict` in the .js files.

PR #541, opened 2017-11-17
Generate Natural Language Classifier with Adapter

glennrfisher:
- [x] Generate service
- [x] Copy tests for use with adapter
- [x] Update tests for generated API
- [x] Add adapter from handwritten to generated API
- [x] Update examples file


examples/natural_language_classifier.v1.js
```diff
@@ -11,12 +11,11 @@ const natural_language_classifier = new NaturalLanguageClassifierV1({
 
 // Creating a classifier
 const params = {
-  language: 'en',
-  name: 'my-classifier',
-  training_data: fs.createReadStream('./resources/weather_data_train')
+  training_data: fs.createReadStream('../test/resources/weather_data_train.csv'),
+  metadata: Buffer.from(JSON.stringify({language: 'en', name: 'my-classifier'}), 'utf8')
```
AmmarD212: I don't think you need call `Buffer.from` here. `JSON.stringify` should do.

PR #546, opened 2017-11-21
Core-SDK Improvements

ammardodin:
**Summary:** 

 - [x] `BaseService` prefers an `api-key` over basic authorization for services which accept an api-key. 
- [x] `BaseService` no longer needs a `constructor` property to access the child class's static `URL` member. This was a hack I did earlier to navigate TypeScript's peculiarities.
- [x] `createRequest` strips `formData` keys with `null/undefined` values, or empty objects since `request` chokes in those scenarios and it's never the right thing to send in a `formData`.
- [x] `createRequest` stringifies any `formData` parameter that is not a file and happens to be an object.
- [x] `createRequest` carries the responsibility of transforming each `formData` parameter that is a file into a request-style object.
- [x] Refactor helper interfaces.
- [x] Refactor VR adapter.

### DIFF
--- a/lib/requestwrapper.ts
+++ b/lib/requestwrapper.ts
@@ -16,8 +16,12 @@
 
 import * as extend from 'extend';
 import * as request from 'request';
-import * as isStream from 'isstream';
-import { getMissingParams, isFileObject } from './helper';
+import {
+  getMissingParams,
+  buildRequestFileObject,
+  isFileParam,
+  isEmptyObject
+} from './helper';
 import { PassThrough as readableStream } from 'stream';
 import { readFileSync } from 'fs';
 import { join } from 'path';
@@ -147,15 +151,7 @@ export function createRequest(parameters, _callback) {
     parameters.defaultOptions,
     parameters.options
   );
-  const path = options.path;
-  const body = options.body; // application/json or text/plain
-  const form = options.form; // application/x-www-form-urlencoded
-  const formData = options.formData; // application/x-www-form-urlencoded
-  const qs = options.qs; // Query parameters
-
-  // Provide a default callback if it doesn't exists
-  const callback =
-    typeof _callback === 'function' ? _callback /* no op */ : function() {};
+  const { path, body, form, formData, qs } = options;
 
   // Missing parameters
   if (parameters.options.requiredParams) {
@@ -174,7 +170,7 @@ export function createRequest(parameters, _callback) {
 
   if (missingParams) {
     if (typeof _callback === 'function') {
-      return callback(missingParams);
+      return _callback(missingParams);
     } else {
       const errorStream = new readableStream();
       setTimeout(function() {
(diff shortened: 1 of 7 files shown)


lib/requestwrapper.ts
```diff
@@ -147,15 +151,7 @@ export function createRequest(parameters, _callback) {
...
-  const path = options.path;
-  const body = options.body; // application/json or text/plain
-  const form = options.form; // application/x-www-form-urlencoded
-  const formData = options.formData; // application/x-www-form-urlencoded
-  const qs = options.qs; // Query parameters
-
-  // Provide a default callback if it doesn't exists
-  const callback =
-    typeof _callback === 'function' ? _callback /* no op */ : function() {};
+  const { path, body, form, formData, qs } = options;
```
germanattanasio: It's beautiful

visual-recognition/v3.ts
```diff
@@ -87,75 +86,84 @@ class VisualRecognitionV3 extends GeneratedVisualRecognitionV3 {
...
     }
+    const _params = extend({}, params, {
+      parameters: this.parseParameters(params)
+    });
+    return super.detectFaces(_params, _callback);
   }
 
   createClassifier(params, callback) {
     const _callback = callback ? callback : () => {};
+    const err = new Error(
```
germanattanasio: How do you handle errors due to missing parameters now?

visual-recognition/v3.ts
```diff
@@ -87,75 +86,84 @@ class VisualRecognitionV3 extends GeneratedVisualRecognitionV3 {
...
-    return Object.keys(_obj).length > 0 ? _obj : null;
+    return Object.keys(_parameters).length > 0 ? _parameters : null;
   }
 
   classify(params, callback) {
-    const _parameters = this.parseParameters(params, callback) || {};
-    // set defaults for classify()
-    if (!_parameters.classifier_ids) {
-      _parameters.classifier_ids = ['default'];
+    const _callback = typeof callback === 'function' ? callback : () => {};
```
germanattanasio: What if you use: 
```js
classify(params, callback = Function.prototype)
```

If `callback` is not a function we should throw an error rather than set a default no-op func. 
What do you think?

visual-recognition/v3.ts
```diff
@@ -87,75 +86,84 @@ class VisualRecognitionV3 extends GeneratedVisualRecognitionV3 {
...
-      classifier_ids: _classifier_ids,
-      owners: _owners,
-      threshold: _threshold
-    };
-    // remove null/undefined keys
-    Object.keys(_obj).forEach(key => {
-      _obj[key] == null && delete _obj[key];
+  }
+
+  private parseParameters(params) {
```
germanattanasio: 🤔

lib/requestwrapper.ts
ammardodin: 🤰

visual-recognition/v3.ts
ammardodin: What's the face for ?

visual-recognition/v3.ts
ammardodin: This is what we're doing in the services methods now.
```
{{#hasRequiredParams}}
    const _callback = (callback) ? callback : () => {};
{{/hasRequiredParams}}
    {{^hasRequiredParams}}
    const _params = (typeof params === 'function' && !callback) ? {} : extend({}, params);
    const _callback = (typeof params === 'function' && !callback) ? params : (callback) ? callback : () => {};
    {{/hasRequiredParams}}
```

PR #556, opened 2017-11-27
Fix formdata bug

ammardodin:


lib/requestwrapper.ts
```diff
@@ -14,8 +14,8 @@
  * limitations under the License.
  */
 
-import * as extend from 'extend';
```
germanattanasio:
```
import extend from 'extend';
```

ammardodin: That doesn't work. The type definition for `extend` is written in such a way that it overloads a bunch of functions. The alternative to `import *` is `import extend = require('extend')`

PR #557, opened 2017-11-27
Generated TTS with Adapter

ammardodin:


text-to-speech/v1-generated.ts
```diff
@@ -0,0 +1,1090 @@
...
+    text: string;
+  }
+
+  /** Constants for the `synthesize` operation. **/
+  export namespace SynthesizeConstants {
+    /** Requested audio format (MIME type) of the audio. You can use this header or the `accept` query parameter to specify the audio format. (For the `audio/l16` format, you can optionally specify `endianness=big-endian` or `endianness=little-endian`; the default is little endian.). **/
+    export enum Accept {
+      BASIC = 'audio/basic',
+      FLAC = 'audio/flac',
+      L16_RATE_NNNN = 'audio/l16;rate=nnnn',
```
germanattanasio: `____` is the rate value. 
e.g. `audio/l16;rate=16000`

ammardodin: The tests are passing, so I am assuming `synthesize()` is functional. As for the user requested audio formats, the user specifies the desired format via either an `accept` query parameter or a `accept2` header parameter.

PR #565, opened 2017-11-30
Final touches [DONT MERGE]

ammardodin:


dialog/v1.js
```diff
@@ -16,18 +16,18 @@
...
+var omit = require('object.omit');
+var isStream = require('isstream');
+var util = require('util');
+var BaseService = require('../lib/base_service').BaseService;
 
 /**
  *
  * @param {Object} options
- * @constructor
+ * @varructor
```
jzhang300: what's a varructor?

jzhang300: oh looks like this is a typo from search and replacing `const`'s

PR #567, opened 2017-11-30
Update npm package

germanattanasio:
* Remove unused dependencies like Karma
 * ignore files from the npm package using `.npmignore`


.gitignore
```diff
@@ -29,3 +29,4 @@ tone-analyzer/*.js
 visual-recognition/*.js
 text-to-speech/*.js
 speech-to-text/*.js
+.nyc_output
```
mamoonraja: .nyc_output/* already exists, we need to get rid of one.

PR #572, opened 2017-12-01
Updates for Beta 2

ammardodin:


germanattanasio:
What is this doing?

ammardodin:
@germanattanasio This gets rid of naming the generated class as `GeneratedClassName`. It adds unit tests for most of the core pieces to bring up code coverage for the `lib` folder to ~90%. And, it moves version dates to the adapters.

PR #577, opened 2017-12-06
Pre Release Clean Up

ammardodin:
* move typings to dependencies
* use `~` in package.json dependency specification
* updated dependencies to latest versions
* remove `recognizeSession`, `recognizeSessionless`, and `observeResult` from generated speech-to-text.


package.json
```diff
@@ -44,48 +44,48 @@
...
+    "@types/extend": "~3.0.0",
+    "@types/file-type": "~5.2.1",
+    "@types/is-stream": "~1.1.0",
+    "@types/node": "~8.0.55",
+    "@types/request": "~2.0.7",
+    "async": "~2.6.0",
+    "buffer-from": "~0.1.1",
     "cookie": "~0.3.1",
-    "csv-stringify": "~1.0.2",
+    "csv-stringify": "~2.0.0",
```
germanattanasio: I'm a little bit worried about this 😱  it's a major release. Can you check if they changed something that may break users trying to use json to create a classifier?

ammardodin: Yeah, it actually broke the `node 4` build. I am reverting this one back to `1.0.2`.

package.json
```diff
@@ -44,48 +44,48 @@
...
+    "eslint-plugin-prettier": "~2.3.1",
+    "jsdoc": "~3.5.5",
+    "memory-fs": "~0.4.1",
+    "mocha": "~4.0.1",
+    "nock": "~9.1.4",
+    "nyc": "~11.4.0",
+    "object.assign": "~4.0.4",
+    "prettier": "~1.9.1",
+    "sinon": "~4.1.2",
+    "typescript": "~2.6.2",
```
ammardodin: Should `typescript` be a dependency and not a dev-dependency ?

PR #587, opened 2017-12-11
Adding missing parameters to speech-to-text

anweshan:
This adds missing parameters to speech-to-text adapter along with adding tests to check that createlanguagemodel, addword, and addwords work


speech-to-text/v1.ts
```diff
@@ -31,7 +31,10 @@ const PARAMS_ALLOWED = [
   'profanity_filter',
   'smart_formatting',
   'customization_id',
-  'speaker_labels'
+  'speaker_labels',
+  'customization_weight',
+  'acoustic_customization_id'
+
```
germanattanasio: remove empty line

PR #593, opened 2017-12-12
Fix Discovery Adapter

### LINKED ISSUE #591: [discovery] error while querying collection using return parameter
While querying a collection if i use return parameter like, `return: ['enriched_text, extracted_metadata']`
the response is 

> TypeError: _params.return.split is not a function . 

This is issue for SDK > 3, but where as for SDK < 3 the response is returned with specified fields.

ammardodin:
Fixes #591 . The adapter should not call `split` on the return parameter. The `requestwrapper` will handle array query parameters.


anweshan:
This looks good

PR #628, opened 2018-02-08
Rename base service version property to serviceVersion

mkistler:
This PR changes the name of the internal `version` property of BaseService to `serviceVersion`.  I believe this is purely an internal change.  All tests pass.

The purpose of this change is to lay the foundation for using 'version' as the parameter name for the "version date" that is passed in to some services.

##### Checklist

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -18,8 +18,6 @@
  * @module watson-developer-cloud
  */
 
-import * as extend from 'extend';
-
 export import AuthorizationV1 = require('./authorization/v1');
 
 export import ConversationV1 = require('./conversation/v1');
@@ -89,10 +87,11 @@ Object.keys(servicesByVersion).forEach(serviceName => {
         );
       }
 
-      const _options = extend({}, options);
-      _options.serviceVersion = options.version;
-      _options.version = options.version_date;
-      return new Service(_options);
+      return new Service({
+        ...options,
+        serviceVersion: options.version,
+        version: options.version_date,
+      });
     }
   });
 });
--- a/index.ts
+++ b/index.ts
@@ -18,6 +18,8 @@
  * @module watson-developer-cloud
  */
 
+import * as extend from 'extend';
+
 export import AuthorizationV1 = require('./authorization/v1');
 
 export import ConversationV1 = require('./conversation/v1');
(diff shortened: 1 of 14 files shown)


index.ts
```diff
@@ -65,16 +65,16 @@ Object.keys(servicesByVersion).forEach(serviceName => {
...
 
-      // previously, AlchemyAPI did not require a version to be specified
+      // previously, AlchemyAPI did not require a serviceVersion to be specified
       if (serviceName.indexOf('alchemy_') === 0) {
-        options.version = 'v1';
+        options.serviceVersion = 'v1';
       }
 
-      const Service = servicesByVersion[serviceName][options.version];
+      const Service = servicesByVersion[serviceName][options.serviceVersion];
```
germanattanasio: This is a breaking change 🔥 

Users should still be able to use

```js
var conversationV1 = watson.conversation({
  version: 'v1',
  version_date: '2017-05-26'
});
```

mkistler: Okay ... I see now.  We documented this style in past releases.  It's good to see that we have removed the doc on this.  I will think about how we can handle that.

mkistler:
I've reworked this change to support the "old style" of service instantiation, but I added a deprecation warning to discourage folks from using it.  I reverted all the changes in the test and examples directory, and all (unchanged) tests pass with this new implementation.

index.ts
```diff
@@ -78,7 +89,10 @@ Object.keys(servicesByVersion).forEach(serviceName => {
         );
       }
 
-      return new Service(options);
+      const _options = extend({}, options);
```
germanattanasio: Did you try with: 
```js
return new Service({
  ...options,
  serviceVersion: options.version,
  version: options.version_date,
});
```

germanattanasio:
This was released as `beta` in `3.0.7`
You can install it by running:
```js
npm install watson-developer-cloud@beta
```

glennrfisher:
Doesn't Natural Language Understanding still need to be updated [here][1] and [here][2]?

As an aside, it can be tricky to follow changes when they're rebased. I think the general recommendation is not to rebase code that's been pushed. (Although there are some nuances, like rebasing a PR into master instead of merging it.)

[1]: https://github.com/watson-developer-cloud/node-sdk/blob/34ae1e74ab55a76a2d2e2665fb75aeefe5b85090/natural-language-understanding/v1.ts#L29

[2]: https://github.com/watson-developer-cloud/node-sdk/blob/34ae1e74ab55a76a2d2e2665fb75aeefe5b85090/natural-language-understanding/v1.ts#L210

[3]: https://www.atlassian.com/git/articles/git-team-workflows-merge-or-rebase

mkistler:
@glennrfisher I believe you are right.  Thanks for catching these.  I guess the tests don't exercise the old constructor for NLU.  But I will fix.

PR #635, opened 2018-02-12
Part 2 of fix for version_date parameter

mkistler:
This PR changes the `version_date` parameter of the generated services to be named `version`.  It also makes a small addition to the "compatibility services" to transfer any value passed in `version_date` to `version` before calling the constructor for the generated service.  This should maintain backward compatibility for existing code, while allowing new code to use `version`.

All the changes in this PR were made by hand, but a PR will be opened for the generator shortly that will generate the SDK following these new conventions.

In this PR, I've left all the tests and examples using `version_date`, to demonstrate that backward compatibility is maintained.  A followup PR will change the tests and examples to use `version` to 
promote this as the preferred way to specify the version-date on service constructors going forward.

##### Checklist


discovery/v1.ts
```diff
@@ -170,7 +177,7 @@ class DiscoveryV1 extends GeneratedDiscoveryV1 {
   }
 
   query(params, callback) {
-    const _params = params || {};
+    let _params = params || {};
```
glennrfisher: Any particular reason this was changed from a `const` to a `let`? It doesn't affect the correctness of the pull request in any way, but since `_params` doesn't change after it is defined, it's nice to express/guarantee that with a `const`.

natural-language-understanding/v1.ts
```diff
@@ -14,653 +14,15 @@
  * limitations under the License.
  */
 
-import * as extend from 'extend';
-import { RequestResponse } from 'request';
-import { createRequest } from '../lib/requestwrapper';
-import { getMissingParams } from '../lib/helper';
-import { BaseService } from '../lib/base_service';
+import GeneratedNaturalLanguageUnderstandingV1 = require('./v1-generated');
```
germanattanasio: Can you use `import {...} from ....`?

discovery/v1.ts
mkistler: Thanks for catching this Glenn.  This was a victim of some git rebasing.  It _should_ be a `const` and I will fix.

natural-language-understanding/v1.ts
mkistler: Yes ... right ... should have done that.  I will fix.

mkistler: Actually ... no ... can't do that.  It fails with
```
natural-language-understanding/v1.ts(17,10): error TS2305: Module '"/Users/mkistler/Projects/watson-node/wdc-node-sdk/natural-language-understanding/v1-generated"' has no exported member 'GeneratedNaturalLanguageUnderstandingV1'.
```

glennrfisher: At some point I think Ammar manually renamed the generated services to have that `Generated` prefix. Probably so he could use the `import { GeneratedService } from './v1-generated'` syntax while avoiding a naming conflict between the generated code and the adapter.

But the Node template generates `Service` instead of `GeneratedService`. So it looks like it got overwritten when more recent code was generated.

Do we have a strong preference one way or the other? I suppose users will be importing the `v1` adapter whether they use the legacy API or the new API—any calls with the new API will just be transparently passed along to the generated code. Since users don't directly import `v1-generated`, we could use whatever name we want there.

My preference, however, would be to continue generating `Service` instead of `GeneratedService` so that the generator is easier to use with external services (e.g. Watson Data Platform, Watson Finance, etc.)

mkistler: This also fails:
```
natural-language-understanding/v1.ts(18,8): error TS1192: Module '"/Users/mkistler/Projects/watson-node/wdc-node-sdk/natural-language-understanding/v1-generated"' has no default export.
```

germanattanasio: I think Glenn is right regarding the `Generated` name. I will merge this and we can fix this issue once we generate the service without `Generated`

PR #636, opened 2018-02-14
Change examples and tests to use version to pass version date

mkistler:
This PR completes the work of changing the parameter for specifying the service version-date from `version_date` to `version`.  Earlier PRs made the changes to the services in a backwards compatible fashion.  This PR updates the examples and tests to use `version` rather than `version_date` except in the "adapter" tests, whose purpose is to test backward compatibility.

This change involved converting a number of the tests to use the "new preferred" method for instantiating services -- necessary to move them from using `version_date` to `version`.

I also removed the versioning info from `auth.example.js` and into the tests, to minimize confusion.

Finally, there are some sections of the code that were reformatted -- mostly statements that had been spread across multiple lines that have been flowed back to a single line. I did not make these changes (intentionally).  I believe these changes were imposed by the (dreaded) pre-commit hook that runs `eslint --fix`.

##### Checklist

- [X] `npm test` passes (tip: `npm run autofix` can correct most style issues)

### DIFF
--- a/examples/browserify/public/client.js
+++ b/examples/browserify/public/client.js
@@ -28,7 +28,7 @@ function getToken() {
 function analyze(token) {
   var toneAnalyzer = new ToneAnalyzerV3({
     token: token,
-    version_date: '2016-05-19'
+    version: '2016-05-19'
   });
   toneAnalyzer.tone(
     {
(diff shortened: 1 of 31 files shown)


germanattanasio:
Thanks @mkistler 👍

PR #638, opened 2018-02-19
Regeneration of services and tslint support

anweshan:
- Updates package.json to use tslint
- Fixes [#644](https://github.com/watson-developer-cloud/node-sdk/issues/644) by using tslint, and updating script on commit to use tslint 
- Regenerates all services using latest version of swagger specification
- Updates tests to remove tests on deprecated methods
- Updates tests to use correct version syntax
- Linting changes to generated code
- Linting changes to non generated code


lib/requestwrapper.ts
```diff
@@ -192,6 +196,7 @@ export function createRequest(parameters, _callback) {
     // Remove non-valid inputs for buildRequestFileObject,
     // i.e things like {contentType: <contentType>}
     Object.keys(formData).forEach(key => {
+      // tslint:disable-next-line:no-unused-expression
```
ammardodin: Why don't we make the logic easier to read and refactor this block that parses the `formData` into it's function to make testing easier? 

Something like this maybe

```
 function parseFormData(formData: Object): Object {
   let _formData = extend({}, formData);
   // cleanup file formData with null/undefined data
   Object.keys(_formData).forEach(key => {
     if (_formData[key].data == null) {
       delete _formData[key];
     }
   });
   // cleanup empty objects
   Object.keys(_formData).forEach(key => {
     if (isEmptyObject(_formData[key])) {
       delete _formData[key];
     }
   });
   // convert file formData parameters to request-style objects
   // and stringify object formData parameters
   Object.keys(_formData).forEach(key => {
     if (isFileParam(_formData[key])) {
       _formData[key] = buildRequestFileObject(_formData[key]);
     } else if (typeof formData[key] === 'object') {
       _formData[key] = JSON.stringify(_formData[key]);
     }
   });
   return _formData;
 }
```

ammardodin: How about we refactor this for readability and place it into a function?

Something like this 

```javascript
function parseFormData(formData: Object): Object {
  let _formData = extend({}, formData);
  // cleanup file formData with null/undefined data
  Object.keys(_formData).forEach(key => {
    if (_formData[key].contentType && formData[key].data == null) {
      delete _formData[key];
    }
  });
  // cleanup empty objects
  Object.keys(_formData).forEach(key => {
    if (isEmptyObject(_formData[key])) {
      delete _formData[key];
    }
  });
  // convert file formData to request-style objects
  // and stringify JSON formData
  Object.keys(_formData).forEach(key => {
    if (isFileParam(_formData[key])) {
      _formData[key] = buildRequestFileObject(_formData[key]);
    } else if (typeof formData[key] === 'object') {
      _formData[key] = JSON.stringify(_formData[key]);
    }
  });
  return _formData;
}
```

anweshan: Refactoring is good, but do we even need this code in the first place? Why does typescript say this code is unused, it is a bit unclear as to its purpose

ammardodin: To understand why we need this code, let's take a look at `createModel` from Language Translator.

This snippet
```javascript
const formData = {
      forced_glossary: {
        data: _params.forced_glossary,
        contentType: 'application/octet-stream'
      },
      parallel_corpus: {
        data: _params.parallel_corpus,
        contentType: 'application/octet-stream'
      },
      monolingual_corpus: {
        data: _params.monolingual_corpus,
        contentType: 'text/plain'
      }
    };
```

That first part of that code is needed to clean up things that looks like this
```javascript
{
  data: undefined,
  contentType: 'application/octet-stream'
}
```
We'll run into this situation for example if the user didn't pass the optional `forced_glossary` when they called `createModel`.

The second part that cleans up empty object literals is needed for the scenario in which we have a JSON formData parameter than end up being passed in as an empty object. The problem here is that the `requests` library will choke 👎  upon receiving those in formData, therefore we need to remove those before passing the request properties to the library. An example of this is the `parameters` parameter in Visual Recognition's `classify`.

The third part has two jobs. First is to Stringify JSON formData parameters before passing them along to `requests`, without doing so `requests` will choke 👎 . Second is responsible for constructing a `requests-style` formData object for each file parameter. This is done so that we uniformly handle how we send files to the services, despite their individual peculiarities around needing filenames or content types to complete the request. If you take a look at `buildRequestFileObject`, you'll see that it will attempt to guess the content type if it's not provided which is often important for services like VR, and it will also try to set a filename if one isn't given, which is important for services like Discovery.

Lastly, the TypeScript linter was issuing a `unused-expression` warning. I believe this is due to the linter being confused about whether some of the expressions in the code have side effects. I think if you rewrite the code with if statements, you shall no longer see those warnings. See [this](https://palantir.github.io/tslint/rules/no-unused-expression/).

package.json
```diff
@@ -48,10 +48,10 @@
     "codecov": "^3.0.0",
     "concat-stream": "~1.6.0",
     "dependency-lint": "~5.0.1",
-    "eslint": "~4.17.0",
+    "eslint": "~4.18.1",
     "eslint-config-google": "~0.9.1",
     "eslint-config-prettier": "~2.9.0",
-    "eslint-plugin-node": "~6.0.0",
+    "eslint-plugin-node": "^6.0.0",
```
germanattanasio: We should keep it as `~` not `^`

package.json
```diff
@@ -61,8 +61,11 @@
     "nock": "^9.1.6",
     "nyc": "^11.4.1",
     "object.assign": "~4.1.0",
-    "prettier": "~1.10.2",
+    "prettier": "1.10.2",
     "sinon": "^4.3.0",
+    "tslint-config-prettier": "^1.9.0",
+    "tslint-eslint-rules": "^5.1.0",
+    "tslint": "^5.9.1",
```
germanattanasio: Let's make sure all our dependencies are `~` instead of `^`

test/unit/test.discovery.v1.js
```diff
@@ -552,7 +552,7 @@ describe('discovery-v1', function() {
             paths.federatedquery +
             '?version=' +
             service.version + // query string params order changed, shouldn't be a problem for the service...
-              '&collection_ids=%5Bcol1-guid%2Ccol2-guid%5D&filter=yesplease&natural_language_query=a%20question%20about%20stuff%20and%20things&count=10&sort=%2Bfield_1%2C-field_2'
+              '&collection_ids=col1-guid%2Ccol2-guid&filter=yesplease&natural_language_query=a%20question%20about%20stuff%20and%20things&count=10&sort=%2Bfield_1%2C-field_2'
```
germanattanasio: Can we check that more than 1 collection id can be query at the same time? I'm worried about this change

lib/requestwrapper.ts
anweshan: This specific code you suggested didn't work and broke tests, but this can be looked into later for improving readability

Issue #655, opened 2018-03-08
The adapter is not forwarding all the parameters to the generated code


germanattanasio:
@mkistler found this issue when trying to help a user with the `tone_chat()` method in Tone Analyzer.

The adapter is currently doing

```js
  tone_chat(params, callback) {
    if (params && params.utterances && params.utterances.utterances) {
      params.utterances = params.utterances.utterances;
    }

    const missingParams = getMissingParams(params, ['utterances']);
    if (missingParams) { return callback(missingParams); }

    const _params: GeneratedToneAnalyzerV3.ToneChatParams = {
      utterances: params.utterances
    };

    return super.toneChat(_params, callback);
  }
```

Because we are not sending the `params` to the generated code but rather building a new `_params` object, we are not allowing new parameters like `content_language` to be added to the service in a backward compatible way.

One of the proposals is to rewrite the method above like:

```js
  tone_chat(params, callback) {
    const missingParams = getMissingParams(params, ['utterances']);
    if (missingParams) { return callback(missingParams); }

    const _params = extend({}, params);
    if (params.utterances.utterances) {
        _params.utterances = params.utterances.utterances;
    }
    return super.toneChat(_params, callback);
  }
```

## Acceptance criteria

- [ ] Should verify that methods are not discarding new parameters that could be added to the service.
- [ ] Should update hand-written methods if a workaround like the one above is needed.

Issue #654, opened 2018-03-08
Display warning message if adapter method signature is different than generated method


germanattanasio:
For example:

In Tone Analyzer, there is a `tone_chat()` but the generated method is `toneChat()`.

We need to display a warning for users using `tone_chat()` to move to `toneChat()`

## Acceptance criteria

- [ ] should display warnings for the methods where the signature is different

PR #656, opened 2018-03-08
Update adapters to log a warning when used

anweshan:
Fixes issues ((#655)) and ((#654))


conversation/v1.ts
```diff
@@ -64,18 +65,17 @@ class ConversationV1 extends GeneratedConversationV1 {
...
-      intent: params.old_intent,
-      new_intent: params.intent,
-      new_description: params.description,
-      new_examples: params.examples
-    };
+    const _params = extend({}, params);
+    _params.intent = params.old_intent;
+    _params.new_intent = params.intent;
+    _params.new_description = params.description;
+    _params.new_examples = params.examples;
```
germanattanasio: The problem is that you will still send the old parameters along with the new parameters. You should not send the old parameters.

PR #667, opened 2018-03-20
Regenerates code to correctly have context parameters as optional

anweshan:
Previously the interface context had required parameters 'conversation_id' and 'system', and the interface counterexample had required parameters 'created' and 'updated', but these should be optional

addresses same issue as:
https://github.com/watson-developer-cloud/python-sdk/pull/414


anweshan:
The travis builds fail because visual_recognition_integration_custom_classifiers does not pass - there is an issue with visual recognition that causes our integration test to fail and time out with an unspecified error, the error is related to using custom classifiers.

Issue #669, opened 2018-03-27
Natural Language Understanding - Removal of constants prevents backward compatibility for version_date from working


chughts:
Because the version constants have been removed, the backward compatibility for version_date for the Natural Language Understanding service does not work. I am using Node v 6.11.2, SDK 3.2.1. 

I have the following code 
````
    var nlu = null,
      serviceSettings = {
        username: username,
        password: password,
        version_date: NaturalLanguageUnderstandingV1.VERSION_DATE_2017_02_27
      };

    if (endpoint) {
      serviceSettings.url = endpoint;
    }

    nlu = new NaturalLanguageUnderstandingV1(serviceSettings);
````

Which in 3.0.x versions of the SDK was ok, but now that the constants have been removed fails with `"Error: Argument error: version was not specified"`

**Comment by anweshan** [comment]
The correct usage is

```
    var nlu = null,
      serviceSettings = {
        username: username,
        password: password,
        version: '2017-02-27',
      };

    if (endpoint) {
      serviceSettings.url = endpoint;
    }

    nlu = new NaturalLanguageUnderstandingV1(serviceSettings);
```

`version_date` has been replaced with `version` and date constants have been removed.

To avoid breaking users we can try to add more backwards compatibility. Currently there is backwards compatibility ensuring that `version_date` still works for users who haven't switched to `version` but not backwards compatibility for the constants.

PR #670, opened 2018-03-27
fix(NLU): Added backwards compatibility for version date constants

anweshan:
Fixes https://github.com/watson-developer-cloud/node-sdk/issues/669


anweshan:
Again test will fail because of visual recognition

germanattanasio:
@anweshan can you add this to the other services too?

anweshan:
As per what I checked it's already in all the other services that formerly had date constants... I can double check if I missed one

PR #672, opened 2018-04-03
Add support for IAM tokens

dpopp07:
Making this PR to show what I've done so far. Would like to get feedback/suggestions on the implementation:

As far as credentials passed into the object constructor, the user would send the API key linked to their IBM Cloud account, referred to in the docs as their "Platform API key". So, I used the variable name `platform_api_key` as the property in the constructor options.

To manage tokens per instance, I'm using the object state to store the access token, refresh token, etc. in `baseService.ts`. To do this, I added a class method called `createRequest` to be a wrapper around the `sendRequest` method in `requestwrapper.ts` (previously called `createRequest`, this naming change decreased the surface area I would need to touch to make this PR). That way, the tokens can be read/written to/from the object state before the request is fired off to the service. The only change this required in the SDKs: changing
```js
return createRequest(parameters, _callback);
// to
return this.createRequest(parameters, _callback);
```

If tokens are unsupported or a `platform_api_key` isn't given (implying the user is using a different method for authentication) `sendRequest` will be called as usual, providing backwards compatibility.

If a service is supporting IAM token authentication, the SDK will define `tokenInfo = {}`. Otherwise, `tokenInfo` will be undefined and that is how the code in `baseService.ts` will determine if it should be using tokens.

I also added a token request function in `helper.ts` and tried to generalize the options in case we ever wanted to support tokens outside of IBM's IAM implementation.

##### Checklist

- [ ] tests are included


germanattanasio:
Hi @dpopp07 thanks for the PR:sunglasses::sunglasses::sunglasses:.

I have a few comments on the approach:

1. IAM vs CF should be determined in the constructor. It won't change during runtime
1. Will `platform_api_key` be confused with the API keys you get in `https://console.bluemix.net/iam/#/apikeys`. We probably want to refer to the IAM apikey for the service instance. Maybe we just use `iam_apikey` or simply `apikey` since that's what they will see in `VCAP_SERVICES` 
    Users may also send us:

    ```json
    { 
      "username": "apikey",
      "apikey": "<IAM APIKEY>"
    }
    ```
    That's completly valid.
1. Users can give us the iam token directly so that we don't have to fetch it. This is the case where they manage the token outside of the SDK
1. Users may give us a [Watson token](https://console.bluemix.net/docs/services/watson/getting-started-tokens.html) which is created from the `Authorization` service
1. Move the functionality to get an IAM token to be like the AUthorization service. a different module rather than a method in the base class. We need to also take into account that the iam url could be different in different environments so we may need to have an `iam_url` parameter in the constructor. I imagine Assistant using the IAM service to get a token.
1. Something we can do to refresh the token is to make it `null` with a `setTimeout()`

dpopp07:
@anweshan No, this PR is in addition to #677.
- That PR dealt with accepting an access token that the user is responsible for managing.
- This PR deals with accepting an API key from the user and managing the token for them (requesting and refreshing the access token as needed).

I am still working on this PR, fixing some things up based on feedback from German and Glenn.

germanattanasio:
@dpopp07 are you going to update this PR?

dpopp07:
@germanattanasio Yes. I'm working on it now, hope to push by this afternoon.

lib/base_service.ts
```diff
@@ -34,6 +35,7 @@ export interface UserOptions {
   username?: string;
   password?: string;
   api_key?: string;
+  platform_api_key?: string;
```
glennrfisher: I suspect you're already working on this change, but I think we settled on something like `iam_apikey` instead.

lib/base_service.ts
```diff
@@ -51,11 +53,20 @@ export interface Credentials {
   username: string;
   password: string;
   api_key: string;
+  platform_api_key?: string;
```
glennrfisher: I suspect you're already working on this change, but I think we settled on something like `iam_apikey` instead.

lib/base_service.ts
```diff
@@ -51,11 +53,20 @@ export interface Credentials {
...
   url: string;
 }
 
+export interface TokenData {
+  access_token?: string;
+  refresh_token?: string;
+  token_type?: string;
+  expires_in?: number;
+  expiration?: number;
+}
```
glennrfisher: Might be worth renaming this to `IAMTokenData`, since other token providers could have different fields.

(For example, [Watson Tokens][1] only return the `access_token` last I checked. There's no information in the response about a `refresh_token` or expiration.)

[1]: https://console.bluemix.net/docs/services/watson/getting-started-tokens.html

glennrfisher: Would it also be reasonable to make some or all of these fields `required` instead of `optional`?

lib/base_service.ts
```diff
@@ -129,12 +145,58 @@ export class BaseService {
...
+
+      // condition 3: the token was retrieved, but has expired
+    } else if (this.tokenIsExpired()) {
+      tokenRequest(this.tokenInfo.refresh_token, (tokenResponse) => {
+        this.setTokenObject(tokenResponse);
+        parameters.defaultOptions.headers.Authorization =
+          `Bearer ${this.tokenInfo.access_token}`;
+        return sendRequest(parameters, _callback);
+        // this 'true' argument specifies that we're refreshing an existing token
+      }, true);
```
glennrfisher: I'm glad you have this comment. I was just wondering, "both of these calls to `tokenRequest` look the same, but one passes the API key and the other passes the refresh token...how does it know?"

It would be pretty easy to miss this, though. I wonder if it would make more sense to move it into the second argument position and pass it in line 175 as well?

`tokenRequest(this._options.iam_apikey, false, (tokenResponse) => { ... })`
`tokenRequest(this.tokenInfo.refresh_token, true, (tokenResponse) => { ... })`

lib/base_service.ts
```diff
@@ -129,12 +145,58 @@ export class BaseService {
...
+
+      // condition 4: the token is good to go, use it as is
+    } else {
+      parameters.defaultOptions.headers.Authorization =
+        `Bearer ${this.tokenInfo.access_token}`;
+      return sendRequest(parameters, _callback);
+    }
+  }
```
glennrfisher: This code block seems pretty specific to IAM tokens. I wonder if there's an opportunity to create a more general credentials manager? Then you could have a file like `IAMTokenManager` that contains the logic for managing IAM tokens, rather than the base service class.

All of this code is internal, though, so that could always be abstracted later.

lib/base_service.ts
```diff
@@ -220,12 +294,14 @@ export class BaseService {
     const _username: string = process.env[`${_name}_USERNAME`] || process.env[`${_nameWithUnderscore}_USERNAME`];
     const _password: string = process.env[`${_name}_PASSWORD`] || process.env[`${_nameWithUnderscore}_PASSWORD`];
     const _apiKey: string = process.env[`${_name}_API_KEY`] || process.env[`${_nameWithUnderscore}_API_KEY`];
+    const _platformApiKey: string = process.env[`${_name}_PLATFORM_API_KEY`] || process.env[`${_nameWithUnderscore}_PLATFORM_API_KEY`];
```
glennrfisher: Just a note that this environment variable should match what's being provided in `VCAP_SERVICES` (see German's pull request -- I think it may be just `apikey` right now).

lib/base_service.ts
```diff
@@ -245,4 +321,49 @@ export class BaseService {
...
+   * @private
+   * @returns {boolean}
+   */
+  private tokenIsExpired(): boolean {
+     // use a buffer to prevent the edge case of the token
+     // expiring before the request could be made
+     // the buffer will be a fraction of the total TTL
+    const fractionOfTtl = 0.8;
+    return this.tokenInfo.expiration * fractionOfTtl
+      < Math.floor(Date.now() / 1000);
```
glennrfisher: Shouldn't this be 80% of `expires_in` instead of 80% of `expiration`? I think 80% of `expiration` would be 80% of the time since the epoch.

I think you want to calculate: `Date.now() > request_time + 0.8 * expires_in`

But since you may not have the time when the token request was made, you could instead calculate:
`Date.now() > expiration - 0.2 * expires_in`

lib/base_service.ts
dpopp07: Yep! Already made this change 👍

lib/base_service.ts
dpopp07: > Might be worth renaming this to IAMTokenData, since other token providers could have different fields.
I think that's a good idea!

> Would it also be reasonable to make some or all of these fields required instead of optional?
I was using optional because of the method we decided on to determine if the API supported tokens from the Swagger, which required the object to be set to `{}`. I am not doing this anymore, so it may make sense for the properties to be required.

lib/base_service.ts
dpopp07: I've actually already changed this to use two different functions: one for requesting a token and one for refreshing. Even though the code is similar, I think it is way more clear. Would like to get other opinions though, once I push.

lib/base_service.ts
dpopp07: Great catch, I must not have been thinking too hard when I switched it from an absolute subtraction to a multiplier.

lib/base_service.ts
dpopp07: Someone correct me if I am wrong but I believe these are for pulling in from standard environment variables, while there is another function (`getCredentialsFromBluemix`) for pulling in the `VCAP_SERVICES` credentials.

lib/base_service.ts
dpopp07: This is a good idea and something I thought about when I first started working on it. At first thought, it seems like it may be more effort than it's worth as it doesn't seem likely that we'll need to support multiple token types in the near future.

I'm going to push the new code I have to be reviewed for now, but I will spend some time thinking about how we might implement this more generally.

dpopp07:
Updated

authorization/v1.ts
```diff
@@ -68,7 +67,7 @@ class AuthorizationV1 extends BaseService {
       },
       defaultOptions: this._options
     };
-    return createRequest(parameters, callback);
```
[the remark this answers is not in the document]
dpopp07: This service is still calling `createRequest`, which now just belongs in the base class. It shouldn't make a difference unless the user is using `iam_apikey` to initialize the constructor.

The alternative is to import `sendRequest` call that directly, if we know we will never use IAM to authenticate this request.

lib/iamTokenManager.ts
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * make a request for an access token
+ * @param {string} apikey - api key used to authenticate the access token request
+ * @param {string} url - IAM token retrieval endpoint
+ * @param {Function} cb - callback function to pass result to
+ * @returns {void}
+ */
+export function requestToken(
+  apikey: string,
```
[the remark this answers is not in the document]
dpopp07: You're saying to nest `iam_apikey` and `iam_url` in a `params` object?

lib/base_service.ts
```diff
@@ -236,13 +304,15 @@ export class BaseService {
     const _apiKey: string = process.env[`${_name}_API_KEY`] || process.env[`${_nameWithUnderscore}_API_KEY`];
     const _url: string = process.env[`${_name}_URL`] || process.env[`${_nameWithUnderscore}_URL`];
     const _accessToken: string = process.env[`${_name}_ACCESS_TOKEN`] || process.env[`${_nameWithUnderscore}_ACCESS_TOKEN`];
+    const _iamApiKey: string = process.env[`${_name}_PLATFORM_API_KEY`] || process.env[`${_nameWithUnderscore}_PLATFORM_API_KEY`];
```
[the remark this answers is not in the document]
dpopp07: Good catch

lib/base_service.ts
```diff
@@ -253,12 +323,60 @@ export class BaseService {
...
+      temp = vcapServices.getCredentials('watson_vision_combined');
     } else {
-      _credentials = vcapServices.getCredentials(vcapServicesName);
+      temp = vcapServices.getCredentials(vcapServicesName);
+    }
+    // convert an iam apikey to use the identifier iam_apikey
+    if (temp.apikey && temp.iam_apikey_name) {
+      temp.iam_apikey = temp.apikey;
     }
+    _credentials = temp;
```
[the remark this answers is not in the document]
dpopp07: Good catch

lib/base_service.ts
```diff
@@ -111,6 +133,7 @@ export class BaseService {
       options,
       _options
     );
+    this.tokenInfo = {} as IamTokenData;
```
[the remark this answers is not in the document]
dpopp07: You mean in the case where a user provides their own, managed, `access_token`. I'm not sure, but I think we should only be using `tokenInfo` when we have all of the token data, when we are managing the token.

The `tokenInfo` object will probably be going away with the next iteration anyways, if the data is going to live in the token manager class.

lib/iamTokenManager.ts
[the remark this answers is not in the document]
germanattanasio: yes

iam-token-manager/v1.ts
```diff
@@ -0,0 +1,198 @@
...
+  private iam_apikey: string; // tslint:disable-line variable-name
+  private userAccessToken: string;
+
+  /**
+   * IAM Token Manager Service
+   *
+   * Retreives, stores, and refreshes IAM tokens.
+   *
+   * @param {Object} options
+   * @param {String} options.iam_apikey
```
germanattanasio: Let's use `camelCase` for variables. I agree with @mkistler that we need to be consistent.

Sorry for asking for `snake_case` :(

iam-token-manager/v1.ts
```diff
@@ -0,0 +1,198 @@
...
+   * 4. If this class is managing tokens and has a valid token stored, send it
+   *
+   * @param {Function} cb - callback function that the token will be passed to
+   */
+  public getToken(cb: Function) {
+    if (this.userAccessToken) {
+      // 1. use user-managed token
+      return cb(null, this.userAccessToken);
+    } else if (!this.tokenInfo.access_token) {
+      // 2. request an initial token
```
germanattanasio: two and tree can be merged.

```js
if (!this.tokenInfo.access_token || this.isTokenExpired()) {
  //...
}
```

iam-token-manager/v1.ts
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  /**
+   * Request an IAM token using an API key.
+   *
+   * @param {string} authHeader - The authorization header to send to the service. Defaults to 'Basic Yng6Yng='.
+   * @param {Function} cb - The callback that handles the response.
+   * @returns {void}
+   */
+  public requestToken(authHeader: string, cb: Function): void {
```
germanattanasio: async calls should always have `params` and `callback`. Synchronous functions can have positional parameters because there will not be a `callback` that always needs to be there so it can be flexible.

iam-token-manager/v1.ts
```diff
@@ -0,0 +1,198 @@
...
+
+  /**
+   * Refresh an IAM token using a refresh token.
+   *
+   * @param {string} authHeader - The authorization header to send to the service. Defaults to 'Basic Yng6Yng='.
+   * @param {Function} cb - The callback that handles the response.
+   * @returns {void}
+   */
+  public refreshToken(authHeader: string, cb: Function) {
+    const authorization = authHeader || 'Basic Yng6Yng=';
```
germanattanasio: I think that basic auth won't change so we can probably hardcode it.

iam-token-manager/v1.ts
```diff
@@ -0,0 +1,198 @@
...
+          'Content-type': 'application/x-www-form-urlencoded',
+          Authorization: authorization
+        },
+        form: {
+          grant_type: 'refresh_token',
+          refresh_token: this.tokenInfo.refresh_token
+        }
+      }
+    };
+    sendRequest(parameters, (error, body, response) => {
```
germanattanasio: write everything in one line.

```
sendRequest(parameters, cb);
```

It will send the response as the third parameter which is not bad actually.
We do it in the SDK already

iam-token-manager/v1.ts
```diff
@@ -0,0 +1,198 @@
...
+   * token expiring before the request could be made.
+   *
+   * The buffer will be a fraction of the total TTL. Using 80%.
+   *
+   * @private
+   * @returns {boolean}
+   */
+  private isTokenExpired(): boolean {
+    const fractionOfTtl = 0.8;
+    const timeToLive = this.tokenInfo.expires_in;
```
germanattanasio: What happens if `expires_in` is not defined?

iam-token-manager/v1.ts
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  /**
+   * Save the response from the IAM service request to the object's state.
+   *
+   * @param {IamTokenData} tokenResponse - Response object from IAM service request
+   * @private
+   * @returns {void}
+   */
+  private saveTokenInfo(tokenResponse: IamTokenData): void {
```
germanattanasio: This is a synchronous request so we can use positional parameters

lib/base_service.ts
```diff
@@ -149,15 +173,36 @@ export class BaseService {
...
+   *
+   * @param {Object} parameters - service request options passed in by user
+   * @param {Function} _callback - callback function to pass the reponse back to
+   * @returns {ReadableStream|undefined}
+   */
+  protected createRequest(parameters, _callback) {
+     if (Boolean(this.tokenManager)) {
+      this.tokenManager.getToken((err, accessToken) => {
+        if (err) {
+          return _callback(err, null);
```
germanattanasio:
```
return callback(err);
```

lib/base_service.ts
```diff
@@ -149,15 +173,36 @@ export class BaseService {
...
+  /**
+   * Wrapper around `sendRequest` that determines whether or not IAM tokens
+   * are being used to authenticate the request. If so, the token is 
+   * retrieved by the token manager.
+   *
+   * @param {Object} parameters - service request options passed in by user
+   * @param {Function} _callback - callback function to pass the reponse back to
+   * @returns {ReadableStream|undefined}
+   */
+  protected createRequest(parameters, _callback) {
```
germanattanasio: can we use `callback` rather `_callback`?

lib/base_service.ts
```diff
@@ -253,12 +299,18 @@ export class BaseService {
    */
   private getCredentialsFromBluemix(vcapServicesName: string): Credentials {
     let _credentials: Credentials;
```
germanattanasio: use `credentials` rather than `_credentials`

lib/base_service.ts
```diff
@@ -235,14 +277,18 @@ export class BaseService {
     const _password: string = process.env[`${_name}_PASSWORD`] || process.env[`${_nameWithUnderscore}_PASSWORD`];
```
germanattanasio: I don't think you need to use underscores here.

iam-token-manager/v1.ts
dpopp07: The separation exists here because for `2` the token is _requested_ but for `3` the token is _refreshed_. While they are pretty similar, I think it is cleaner to leave them separate. Let me know what you think

iam-token-manager/v1.ts
dpopp07: This function should never be called if `expires_in` is undefined. `expires_in` gets set at the same time as `access_token`, so the check for `!this.tokenInfo.accessToken` should trigger a request if they aren't defined.

iam-token-manager/v1.ts
dpopp07: The IAM service returns a single object that I am passing into this function to be set to the object state. It seems unnecessary to me to destruct the object into variables, pass them in as positional parameters, and re-construct the object to save in the state. What do you think?

iam-token-manager/v1.ts
germanattanasio: my bad.

iam-token-manager/v1.ts
germanattanasio: I was just giving an example of when to you positional parameters. You method was fine 😋

lpatino10:
@ehdsouza brought up a good point: what about the case where the access token is expired, we try to refresh it, but the refresh token is expired as well?

It sounds like we might need a fallback somewhere to request a new token using the API key in this case.

dpopp07:
I could be wrong, but I don't believe refresh tokens ever expire. They can only be revoked

iam-token-manager/v1.ts
```diff
@@ -0,0 +1,190 @@
...
+  iamUrl?: string;
+}
+
+export interface IamTokenData {
+  access_token: string;
+  refresh_token: string;
+  token_type: string;
+  expires_in: number;
+  expiration: number;
+}
```
glennrfisher: I suspect you've made a conscious decision. But it seems a little odd that `Options` uses `camelCase` while `IamTokenData` uses `snake_case`.

iam-token-manager/v1.ts
```diff
@@ -0,0 +1,190 @@
...
+      // 3. refresh a token
+      this.refreshToken((err, tokenResponse) => {
+        this.saveTokenInfo(tokenResponse);
+        return cb(err, this.tokenInfo.access_token);
+      });
+    } else {
+      // 4. use valid managed token
+      return cb(null, this.tokenInfo.access_token);
+    }
+  }
```
glennrfisher: I like the helper functions that you used here. I think they make it very clear what's going on, even though this logic is relatively complex with multiple branches.

In particular, the use of `isTokenExpired`, `requestToken`, `refreshToken`, and `saveTokenInfo` is on point. 👌

iam-token-manager/v1.ts
```diff
@@ -0,0 +1,190 @@
...
+   * @returns {void}
+   */
+  public requestToken(cb: Function): void {
+    const parameters = {
+      options: {
+        url: this.url,
+        method: 'POST',
+        headers: {
+          'Content-type': 'application/x-www-form-urlencoded',
+          Authorization: 'Basic Yng6Yng='
```
glennrfisher: Could this cause any issues since it's hardcoded? I know you mentioned that it acts like a secret -- you need to pass the same authorization header when refreshing a token.

Since the SDK should be managing any token that is requested/refreshed (and is ignoring the case where users provide a refresh key and secret...at least for now), it seems safe to use this hardcoded value. But just want to check.

iam-token-manager/v1.ts
```diff
@@ -0,0 +1,190 @@
...
+   * The access token should be valid and not yet expired.
+   *
+   * By using this method, you accept responsibility for managing the
+   * access token yourself. You must set a new access token before this
+   * one expires. Failing to do so will result in authentication errors
+   * after this token expires.
+   *
+   * @param {string} iamAccessToken - A valid, non-expired IAM access token
+   * @returns {void}
+   */
```
glennrfisher: Hey, nice comment. ;)

iam-token-manager/v1.ts
glennrfisher: Might be straightforward to add a defensive check, just in case:

```node
if (!this.tokenInfo.expires_in || !this.tokenInfo.expiration) return true;
```

iam-token-manager/v1.ts
```diff
@@ -0,0 +1,190 @@
...
+   *
+   * @private
+   * @returns {boolean}
+   */
+  private isTokenExpired(): boolean {
+    const fractionOfTtl = 0.8;
+    const timeToLive = this.tokenInfo.expires_in;
+    const expireTime = this.tokenInfo.expiration;
+    return expireTime - (timeToLive * (1.0 - fractionOfTtl))
+      < Math.floor(Date.now() / 1000);
```
glennrfisher: This certainly isn't necessary, but it might help to add a couple of intermediate, explanatory variables here:

```node
const expireTime = this.tokenInfo.expiration;
const refreshTime = expireTime - (timeToLive * (1.0 - fractionOfTtl));
const currentTime = Math.floor(Date.now() / 1000);
return refreshTime < currentTime;
```

lib/base_service.ts
```diff
@@ -230,19 +272,23 @@ export class BaseService {
...
-      api_key: _apiKey,
-      url: _url,
-      access_token: _accessToken
+      username,
+      password,
+      api_key: apiKey,
+      url,
+      iam_access_token: iamAccessToken,
+      iam_apikey: iamApiKey,
+      iam_url: iamUrl
```
glennrfisher: Nice. I like the move away from `_` variables.

lib/base_service.ts
```diff
@@ -252,13 +298,19 @@ export class BaseService {
...
+      temp = vcapServices.getCredentials(vcapServicesName);
     }
-    return _credentials;
+    // convert an iam apikey to use the identifier iam_apikey
+    if (temp.apikey && temp.iam_apikey_name) {
+      temp.iam_apikey = temp.apikey;
+      delete temp.apikey;
+    }
+    credentials = temp;
+    return credentials;
```
glennrfisher: Was a little confused to see `temp` used as the variable name here. Although I can see why, given that `credentials` and `vcapServices` are both already defined.

test/unit/test.iam_token_manager.js
```diff
@@ -0,0 +1,126 @@
...
+    instance.setAccessToken(accessToken);
+
+    instance.getToken(function(err, token) {
+      assert.equal(token, accessToken);
+      assert.equal(requestSpy.notCalled, true);
+      assert.equal(refreshSpy.notCalled, true);
+      done();
+    });
+  });
+});
```
glennrfisher: Can always appreciate a good test set! 🍻

assistant/v1.ts
```diff
@@ -113,7 +112,7 @@ class AssistantV1 extends BaseService {
         }
       })
     };
-    return createRequest(parameters, _callback);
+    return this.createRequest(parameters, _callback);
```
glennrfisher: Is `this` still required, even though `createRequest` is now defined by the superclass instead of an import?

lpatino10:
Here's the response I got regarding refresh tokens:

screen shot 2018-04-18 at 9 32 26 am

dpopp07:
@lpatino10 This is good info, thanks for asking and for sharing that! The issue is, at least in the Node SDK, it would be non-trivial to catch an error, check if it has that message, and make a request for a new token. It may even require some rewiring of the architecture. Not sure, but that's my instinct.

I think a simpler but still appropriate solution (again, for the Node SDK) would be to automatically refresh the token after something like 7 days. An amount of time that is significantly less the refresh token's time-to-live but still long enough that it's not a performance concern. I don't think one extra request per week as a fail-safe is unreasonable. What do you think?

iam-token-manager/v1.ts
dpopp07: A conscious decision, yes, because the IAM service response object had exactly those keys. It is designed to match that. I should maybe leave a comment because you aren't the first person to bring it up!

iam-token-manager/v1.ts
dpopp07: It should not cause any issues in the code. The only downside is the use case where a user is using this token manager class on it's own - they would not be able to specify a custom authorization header. Everything should still work, they just would be stuck with the auth header that we provide.

lpatino10:
Yeah that sounds pretty reasonable to me.

dpopp07:
Cool. I realized that I can't really do a proactive refresh, but I will add a condition to request a new token using the API key if it's been 7 days since the access token expired. That will be included in my next push.

lpatino10:
Another case that we might need to add in: handling users passing in just a refresh token.

According to @glennrfisher's last comment [here](https://github.ibm.com/Watson/developer-experience/issues/3265), that is something we should support. If that's the case, we'll need to tweak the logic in `getToken()` to check for the following case:

```javascript
// refresh and get new tokenInfo
if (!this.tokenInfo.access_token && this.tokenInfo.refresh_token) {

}
```

dpopp07:
That was something that we initially discussed but while working on it, I discovered that it would require more code to be able to handle that case. The reason is that whatever authorization header was used for the initial token request needs to be the same for the refresh, so we would need their header and would need to manage them both, etc. Not impossible but probably more complication that it's worth.

German told me not to worry about it for now, so we aren't going to support that for the time being.

PR #674, opened 2018-04-04
feat(visual recognition & nlc): vr: new coreML endpoint nlc: new classifyCollection endpoint

anweshan:
Adds support for two new endpoints and adds unit tests for both of them

### DIFF
--- a/test/unit/test.adapter.natural_language_classifier.v1.js
+++ b/test/unit/test.adapter.natural_language_classifier.v1.js
@@ -5,6 +5,7 @@ const watson = require('../../index');
 const nock = require('nock');
 const fs = require('fs');
 const extend = require('extend');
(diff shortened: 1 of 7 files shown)


test/unit/test.adapter.natural_language_classifier.v1.js
```diff
@@ -159,4 +170,13 @@ describe('natural_language_classifer', function() {
     params.training_data = [{ json: 'bad' }];
     natural_language_classifier.create(params, invalidFormatParameter);
   });
+
+  it('should classify collection', function() {
```
germanattanasio: can we assert that the body, url, and query parameters are correct?

germanattanasio:
Speech to text test is failing.

PR #677, opened 2018-04-06
accept access_token

dpopp07:
#### New Features
- Allow the user to pass an `access_token` into the constructor
- Allow the user to reset this token with a setter method when the token expires and they get a new one
- When a token is passed in the constructor or setter method, the authentication header is automatically set

I also removed the specific error handling that prints the credentials the SDK expected. The way this is implemented would make it too complicated to evolve as we accept multiple styles of authentication. I think it is cleaner and just as informative to tell the user their credentials are wrong and point them to the documentation. Let me know what you think about that.

I also noticed that if you pass in something that would be a valid credential (like an `api_key`) to a service that doesn't support that (like `conversation`), the credentials check passes and the service responds with a 401 and the message `Unauthorized`. I made an edit in `requestwrapper.ts` to print a better error message if a 401 is received by the SDK.

I edited the tests accordingly. All of the tests that pass in `master` are also passing in this branch.

##### Checklist

- [ ] tests are included

##### New version_date Checklist

- [ ] A new constant is avaliable with the version_date - [example](https://github.com/watson-developer-cloud/node-sdk/blob/d1418ac2f9774194aaff0c8bd80f0d3722beef72/conversation/v1.js#L77)
- [ ] The new constant has a comment that summarizes the changes and/or links to relevant doc pages
- [ ] Any older version_date constants remain intact
- [ ] The error message thrown if the service is created without a version_date indicates the new version_date constant
- [ ] The example in the README includes the new version_date constant
- [ ] Any relevant code in the examples/ folder has been updated to use the new version_date constant
- [ ] Most tests are updated to the new version_date
- [ ] 1-2 new tests are added that use the old version_date (optional, but preferred)

### DIFF
--- a/lib/base_service.ts
+++ b/lib/base_service.ts
@@ -141,17 +141,21 @@ export class BaseService {
   }
 
   /**
-   * Set an IAM access token to authenticate the service with
-   * The SDK will assume this token is valid and that this method
-   * will be used to set a new token upon expiration.
+   * Set an IAM access token to use when authenticating with the service.
+   * The access token should be valid and not yet expired.
    *
-   * @param {string} accessToken - Valid IAM access token
+   * By using this method, you accept responsibility for managing the
+   * access token yourself. You must set a new access token before this
+   * one expires. Failing to do so will result in authentication errors
+   * after this token expires.
+   *
+   * @param {string} access_token - A valid, non-expired IAM access token
    * @returns {void}
    */
-  public setAccessToken(accessToken: string) {
-    this._options.access_token = accessToken;
+  public setAccessToken(access_token: string) { // tslint:disable-line variable-name
+    this._options.access_token = access_token;
     this._options.headers = this._options.headers || {};
-    this._options.headers.Authorization = `Bearer ${accessToken}`;
+    this._options.headers.Authorization = `Bearer ${access_token}`;
   }
 
   /**
--- a/lib/base_service.ts
+++ b/lib/base_service.ts
@@ -38,6 +38,7 @@ export interface UserOptions {
   use_unauthenticated?: boolean;
   headers?: HeaderOptions;
   token?: string;
+  access_token?: string;
 }
 
 export interface BaseServiceOptions extends UserOptions {
@@ -52,18 +53,19 @@ export interface Credentials {
   password: string;
   api_key: string;
   url: string;
+  access_token: string;
 }
 
 function hasCredentials(obj: any): boolean {
-  return obj && ((obj.username && obj.password) || obj.api_key);
+  return obj && ((obj.username && obj.password) || obj.api_key || obj.access_token);
 }
 
 function hasBasicCredentials(obj: any): boolean {
   return obj && obj.username && obj.password;
 }
 
-function acceptsApiKey(name: string): boolean {
-  return name === 'visual_recognition';
+function hasAccessToken(obj: any): boolean {
+  return obj && obj.access_token;
 }
 
 export class BaseService {
@@ -132,7 +134,24 @@ export class BaseService {
     if (this._options.url) {
       _credentials.url = this._options.url;
     }
+    if (this._options.access_token) {
+      _credentials.access_token = this._options.access_token;
+    }
     return _credentials;
+  }
+
+  /**
+   * Set an IAM access token to authenticate the service with
+   * The SDK will assume this token is valid and that this method
+   * will be used to set a new token upon expiration.
+   *
+   * @param {string} accessToken - Valid IAM access token
+   * @returns {void}
+   */
+  public setAccessToken(accessToken: string) {
+    this._options.access_token = accessToken;
+    this._options.headers = this._options.headers || {};
+    this._options.headers.Authorization = `Bearer ${accessToken}`;
   }
 
   /**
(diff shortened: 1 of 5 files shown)


germanattanasio:
It looks good. Take into account that some of our services may require user api_key or `username` and `password` at the same time as with Visual Recognition. It only uses `api_key` in dedicated environments.

germanattanasio:
Can you add a test for `setAccessToken()`?

lib/base_service.ts
```diff
@@ -52,18 +53,19 @@ export interface Credentials {
...
 function hasCredentials(obj: any): boolean {
-  return obj && ((obj.username && obj.password) || obj.api_key);
+  return obj && ((obj.username && obj.password) || obj.api_key || obj.access_token);
 }
 
 function hasBasicCredentials(obj: any): boolean {
   return obj && obj.username && obj.password;
 }
 
-function acceptsApiKey(name: string): boolean {
```
anweshan: why does this need to be removed? im wondering if this reduces functionality in some cases for when we check visrec

dpopp07: That function was only being used to determine if an error statement should say "pass in an api_key". I removed that error statement because when supporting more authentication methods (api_key, iam_api_key, access_token, refresh_token, etc) it will get too complicated to be worth printing specific error statements. So I removed that function as well

dpopp07:
@germanattanasio Yes, I will add a test. Good idea.

@anweshan Not sure if I understand you correctly. You are talking about accessing environment variables in the SDK? I added this line in case the user loaded the access_token as an environment variable (Line 234 in base_service.ts):
```js
const _accessToken: string = process.env[`${_name}_ACCESS_TOKEN`] || process.env[`${_nameWithUnderscore}_ACCESS_TOKEN`];
```

dpopp07:
@germanattanasio Tests added for both
1) Setting correct header when `access_token` is passed into constructor
2) Setting correct header with `setAccessToken()`

lib/base_service.ts
```diff
@@ -52,18 +53,19 @@ export interface Credentials {
   password: string;
   api_key: string;
   url: string;
+  access_token: string;
```
glennrfisher: I know you were following the pattern here, but this interface doesn't quite seem right. It seems odd that all of these values are required instead of optional.

lib/base_service.ts
```diff
@@ -132,9 +134,26 @@ export class BaseService {
...
   }
 
+  /**
+   * Set an IAM access token to authenticate the service with
+   * The SDK will assume this token is valid and that this method
+   * will be used to set a new token upon expiration.
+   *
+   * @param {string} accessToken - Valid IAM access token
+   * @returns {void}
+   */
```
glennrfisher: I think this comment could be improved a bit.

```node
/**
 * Set an IAM access token to use when authenticating with the service.
 * The access token should be valid and not yet expired.
 *
 * By using this method, you accept responsibility for managing the
 * access token yourself. You must set a new access token before this
 * one expires. Failing to do so will result in authentication errors
 * after this token expires.
 *
 * @param {string} accessToken - A valid, non-expired IAM access token.
 */

File: lib/base_service.ts [hunk 4]
```diff
@@ -132,9 +134,26 @@ export class BaseService {
...
+  public setAccessToken(accessToken: string) {
```
glennrfisher: Should this be `accessToken` or `access_token`? I don't have a preference, but want to make sure the API is consistent.

dpopp07: `tslint` yells at you if an argument name is not camel case, which is why I used `accessToken`. Also, while it is public facing, it takes in a pure string and the user does not have to identify it with `accessToken`. The constructor takes in an object and requires the identifier `access_token`. So while the argument names are not consistent, I would say the API technically is.

That being said, the difference would be reflected in the documentation and I am happy to change the arg name and add a `tslint-disable-line` comment if that would be preferred.

lib/base_service.ts
dpopp07: I agree. With different methods of authentication allowed, it doesn't seem to fit the general case to make these all required. That being said, is has not caused any type-issues despite never including all of these.

@germanattanasio or @anweshan can you comment on this?

ammardodin: @glennrfisher you're right that it looks odd 🥇 . I think it still works because although those all/some of those keys might have null/undefined values, as long as they're still keys on the object, Typescript doesn't complain ? That's my guess from looking at the methods that return `Credentials`.

anweshan: It hasn't caused any type issues because we are never calling the Credentials interface before it has these values. We instead call things like getCredentials that return an Credentials interface with all these options, which are blank if nothing was there. And then we have methods that call upon getCredentials.username to see if there was a username provided, etc. If the parameters are optional, then I think that when getCredentials() creates a new Credentials interface, it then won't have all the parameters like username, which could lead to exceptions calling for getCredentials.username etc. when not provided.

dpopp07: What do you mean by blank? I had not looked into this much so I am testing out the current behavior. The following code:
```js
const conversation = new ConversationV1({
  url: myUrl,
  version: '2017-05-26',
  access_token: token
});

const credentials = conversation.getCredentials();
console.log(JSON.stringify(credentials, null, 2));
```
produces the output:
```
{
  "url": "<the url>",
  "access_token": "<the token>"
}
```
`credentials.username` returns `undefined`, the same as `credentials.foobar` or any other key that doesn't exist on the object. I am not understanding what the type-safety around requiring all of those keys is granting us.

lib/base_service.ts
germanattanasio: I agree with @glennrfisher. You can disable the rule for that line.

lib/base_service.ts
anweshan: Okay, in that case you are right. I wasn't sure if in TS it will cause an error calling things like `credentials.fodder` in our code, but it does not. In that case it really doesn't matter if the keys are required or not required, since we only call the interface through `getCredentials()` rather than the interface itself...so you can change it to be optional for more clarity, although it won't affect the code itself.

lib/base_service.ts
mkistler: The style rules for Javascript say that we should be using [camel case for naming objects, functions, and instances](https://github.com/airbnb/javascript#naming--camelCase).  Snake case violates the style guide.

That said, I think we have not been doing a good job of generating SDKs that follow this style, so I'm not sure this is worth arguing over.

lib/base_service.ts
dpopp07: Okay, that makes sense. Thanks. This PR is merged anyways but I might make that change in the next

PR #679, opened 2018-04-13
Split off token logic

dpopp07:
PR for comparison against the other implementation.

This implementation involves a new class, `IamTokenManager`, that stores the data and handles the token-managing logic.

Still need to add tests for this one.


iam-token-manager/v1.ts
```diff
@@ -0,0 +1,190 @@
...
+   * @param {String} options.iam_apikey
+   * @param {String} options.access_token
+   * @param {String} options.iam_url - url of the iam api to retrieve tokens from
+   * @constructor
+   */
+  constructor(options: Options) {
+    this.url = options.iam_url || 'https://iam.ng.bluemix.net/identity/token';
+    this.tokenInfo = {} as IamTokenData;
+    if (options.iam_apikey) {
+      this.apikey = options.iam_apikey;
```
germanattanasio: call it iam_apikey here not `apikey`

iam-token-manager/v1.ts
```diff
@@ -0,0 +1,190 @@
...
+
+  /**
+   * This function sends an access token back through a callback. The source of the token
+   * is determined by the following logic:
+   * 1. If user provides their own managed access token, assume it is valid and send it
+   * 2. If this class is managing tokens and does not yet have one, make a request for one
+   * 3. If this class is managing tokens and the token has expired, refresh it
+   * 4. If this class is managing tokens and has a valid token stored, send it
+   */
+  public getManagedToken(cb: Function) {
```
germanattanasio: `getToken()`

iam-token-manager/v1.ts
```diff
@@ -0,0 +1,190 @@
...
+        'Content-type': contentType,
+        Authorization: authorization
+      },
+      form: {
+        grant_type: grantType,
+        apikey: this.apikey,
+        response_type: responseType
+      }
+    };
+    request.post(requestOptions, (err, res, body) => {
```
germanattanasio: Can we make this method similar to the methods in our services where we build an object and call `requestwrapper`?

iam-token-manager/v1.ts
```diff
@@ -0,0 +1,190 @@
...
+
+  public setAccessToken(access_token: string): void { // tslint:disable-line variable-name
+    this.userAccessToken = access_token;
+  }
+
+  private needToRequestToken() {
+    return !this.tokenInfo.access_token;
+  }
+
+  private tokenIsExpired(): boolean {
```
germanattanasio: `isTokenExpired()`

iam-token-manager/v1.ts
```diff
@@ -0,0 +1,190 @@
...
+    // use a buffer to prevent the edge case of the token
```
germanattanasio: Can you move this comment as a method description?

iam-token-manager/v1.ts
```diff
@@ -0,0 +1,190 @@
...
+        'Content-type': contentType,
+        Authorization: authorization
+      },
+      form: {
+        grant_type: 'refresh_token',
+        refresh_token: this.tokenInfo.refresh_token
+      }
+    };
+
+    request.post(requestOptions, (err, res, body) => {
```
germanattanasio: you don't need to check for errors if you reuse `createRequest` from `requestwrapper`

lib/base_service.ts
```diff
@@ -342,41 +317,8 @@ export class BaseService {
...
-    // use a buffer to prevent the edge case of the token
-    // expiring before the request could be made
-    // the buffer will be a fraction of the total TTL
-    const fractionOfTtl = 0.8;
-    const timeToLive = this.tokenInfo.expires_in;
-    const expireTime = this.tokenInfo.expiration;
-    return expireTime - (timeToLive * (1.0 - fractionOfTtl))
-      < Math.floor(Date.now() / 1000);
+  private usingTokenManager(): boolean {
+    const result = Boolean(this.tokenManager);
```
germanattanasio: This can be one line

iam-token-manager/v1.js
```diff
@@ -0,0 +1,159 @@
+"use strict";
```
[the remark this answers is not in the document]
dpopp07: oops, good catch

iam-token-manager/v1.ts
dpopp07: yep, good call

iam-token-manager/v1.ts
dpopp07: Yes, I'll look into that

iam-token-manager/v1.ts
```diff
@@ -0,0 +1,190 @@
...
+  expires_in: number;
+  expiration: number;
+}
+
+export interface TokenRequestParams {
+  contentType?: string;
+  authHeader?: string;
+  grantType?: string;
+  responseType?: string;
+}
```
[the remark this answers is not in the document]
dpopp07: I completely agree with keeping consistent style. However, the interface for `IamTokenData` is based on the response object from the IAM service. Those parameters are the keys in the response object. I was keeping those consistent so a simple `extend` would be sufficient to copy the response data to a local variable.

iam-token-manager/v1.ts
```diff
@@ -0,0 +1,190 @@
...
+
+  public requestToken(params: TokenRequestParams, cb: Function): void {
+    const contentType = params.contentType ||
+      'application/x-www-form-urlencoded';
+
+    const grantType = params.grantType ||
+      'urn:ibm:params:oauth:grant-type:apikey';
+
+    const authorization = params.authHeader || 'Basic Yng6Yng=';
+    const responseType = params.responseType || 'cloud_iam';
```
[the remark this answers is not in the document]
dpopp07: Currently, there is no use case for this. I was just trying to make it easier to generalize in the future, should a use case arise. I am good to remove this though and just use the default parameters.

iam-token-manager/v1.ts
```diff
@@ -0,0 +1,190 @@
...
+      } else if (res.statusCode < 200 || res.statusCode >= 300) {
+        throw new Error(JSON.stringify(res, null, 2));
+      } else {
+        const tokenResponse = JSON.parse(body);
+        cb(tokenResponse);
+      }
+    });
+  }
+
+  public setAccessToken(access_token: string): void { // tslint:disable-line variable-name
```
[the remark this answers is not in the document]
dpopp07: This is another issue of consistency. The constructor options all use snake_case and so I made the `access_token` parameter snake_case to be consistent. Then we have this function, which is public, so German and Glenn asked me to keep the naming of the parameters consistent. Hence, using snake_case.

dpopp07:
Since we seem to be in agreement that the logic can be in another file, I going to merge this PR into the other one and we can move forward reviewing only that PR.

PR #717, opened 2018-06-08
Fix Speech to Text WebSockets recognize

dpopp07:
This PR addresses the issues users were having with STT, specifically using streaming operations with IAM authentication.

Authenticating with an IAM API key requires requests to be made in order to get an access token. These request interrupt the stream and cause the streaming operation to fail. My solution for this is to give users an option (an option, but it's required for IAM and streams to work) to pre-authenticate and receive a `ready` flag that tells them they are good to proceed with the request. It is not the cleanest solution but I have not thought of a better way.

What this function does is make any IAM requests, if necessary. When the user makes their service request, the code will see that an access token is already stored and will launch the request without any delays that would interrupt the stream.

This line:
```js
const authHeader = { Authorization: 'Bearer ' + token };
this._options.headers = extend(authHeader, this._options.headers);
```

is for `createRecognizeStream`, which looks for an authorization header in `this._options.headers`.

An example of using this method:
```js
stt.preAuthenticate(function(ready) {
  if (!ready) {
    return;
  }
  stt.recognize(params, function(err, res) {
    if (err) {
      console.log(err);
    } else {
      console.log(JSON.stringify(res, null, 2));
    }
  });
});
```

We can definitely change the name too. Happy to take any suggestions.

TODO:
- [ ] Run tests
- [ ] Add a section about this in the README


mediumTaj:
Any reason tests are failing?

anweshan:
@germanattanasio Yes, this solution has been tested for STT with an instance in SYD region

germanattanasio:
If this works for CF, RC and in Sydney and Dallas then we just need to make sure we do the `preAuthenticate()` as part of the service call.

dpopp07:
@germanattanasio I agree that it would be better if the user didn't have to call `preAuthenticate` but I'm not sure that it's possible. That is essentially what we are already doing - when the user makes a service call, we run it through the IAM logic and that will send off a request to get a token if we don't already have one. The problem is, when a user requests a _stream_, this async request we make for a token is interrupting the stream (as far as I know).

The `preAuthenticate` method is just ensuring that when a user makes a service call, the IAM logic is forced through the conditional branch that gets the access token from a stored variable rather than another request, thereby allowing the stream to complete without being interrupted.

If you have any ideas, please let me know. I'm certainly not an expert on streams, so there may be a way to make it work. I will continue thinking about it too, but to my current knowledge, I don't think that there is a solution that allows everything to happen behind the scenes.

jsstylos:
Conditionally call `preAuthenticate` when creating recognize stream, if it hasn't been called recently? (Presumably this is what users would have to do in their own code.) Call `preAuthenticate` when creating the service instance?

germanattanasio:
If the user sends a valid token and everything works then the `preAuthenticate()` should only be a call to get a token and then use that token.

dpopp07:
We can't call it when creating the service because constructors must be synchronous in Node.

The problem is that, to my current knowledge, there can be no async operation between the user making the stream request and the SDK making the actual service call. It will only work if there is already a token stored, which `preAuthenticate` enforces. Putting that logic after the user's request is no different than the current IAM logic and will break if a token is not already stored.

I agree that this is unfortunate and I would much prefer a different solution. I am going to spend time today trying to figure out if that is actually the case or if there is some streaming magic we can do to get it to work. But as of right now, I don't think it is possible. I will keep you updated.

germanattanasio:
@dpopp07 can you wrap the stream in another stream that will do the async work? or maybe do the async work before sending the first bite of audio

dpopp07:
Possibly. I'm looking into transform streams right now which may be able to do what we want. Not sure how much I'll have to change the existing stream implementation yet. I'm trying to get the theory down first. I am starting to think it might be possible though, which is good

dpopp07:
UPDATE:

I was able to get this working internally. It is not the cleanest solution but it works and requires no change from the user's perspective. Two different solutions were required for `createRecognizeStream` and `recognize`. I would like feedback on the implementation. I also don't know that I'm happy with the name `preAuthenticate` so any suggestions there are welcome also

lib/base_service.ts
```diff
@@ -192,18 +191,42 @@ export class BaseService {
...
+   *
+   * @param {Function} callback - callback function to return flow of execution
+   *
+   * @returns {void}
+   */
+  protected preAuthenticate(callback): void {
+     if (Boolean(this.tokenManager)) {
+      return this.tokenManager.getToken((err, token) => {
+        if (err) {
+          callback();
```
germanattanasio: the callback should bubble up the error

```js
return callback(err);
```

lib/base_service.ts
```diff
@@ -192,18 +191,42 @@ export class BaseService {
...
+        }
+        const authHeader = { Authorization: 'Bearer ' + token };
+        this._options.headers = extend(authHeader, this._options.headers);
+        callback();
```
germanattanasio:
```
callback(null)
```

lib/base_service.ts
```diff
@@ -192,18 +191,42 @@ export class BaseService {
...
+      });
+    } else {
+      callback();
+    }
+  }
+
```
germanattanasio: why do you do this in a method? I don't like having a method that changes a private variable like `_options()`

lib/recognize-stream.ts
```diff
@@ -398,6 +402,12 @@ class RecognizeStream extends Duplex {
       // can't send any more data after the stop message (although this shouldn't happen normally...)
       return;
     }
+
+    // if using iam, we need to authenticate the first time `_write` is called
+    if (!this.authenticated) {
+      return this.setAuthorizationHeaderToken(callback);
```
germanattanasio: If you call `_write()` and you don't write it to the stream you may lose your first bytes... I think you need to have a callback to check if there is a token and then operates like it used to do.

You can't call the same callback more than once and/or ignore the first bytes.

speech-to-text/v1.ts
```diff
@@ -510,12 +522,15 @@ class SpeechToTextV1 extends GeneratedSpeechToTextV1 {
...
       defaultOptions: this._options
     };
-    return params.audio
-      .on('response', (response) => {
-        // Replace content-type
-        response.headers['content-type'] = params.content_type;
-      })
-      .pipe(this.createRequest(parameters, callback));
+
+    this.preAuthenticate(() => {
```
germanattanasio: You need to check if there is an error or not

lib/base_service.ts
dpopp07: I actually didn't mean to leave that in there, that was mostly  to account for `createRecognizeStream`, which now has a different solution. There is still an issue of the way `recognize` handles headers. It puts them in `parameters.options` rather than `parameters.defaultOptions` like all the generated methods.

This leads to the code crashing when it gets here:
```js
        parameters.defaultOptions.headers.Authorization =
          `Bearer ${accessToken}`;
```

Can we change `recognize` to put the headers in `defaultOptions` to stay consistent? Otherwise, I will add logic like this to `createRequest` in the base service:
```js
if (parameters.defaultOptions.headers) {
  parameters.defaultOptions.headers.Authorization = `Bearer ${accessToken}`;
} else {
  parameters.options.headers.Authorization = `Bearer ${accessToken}`;
}
```

Let me know what you think @germanattanasio

germanattanasio: I don't have a preference here so I will let you decide

dpopp07:
All changes addressed. Please provide any more feedback

germanattanasio:
@dpopp07 this looks better. Did you run some tests with a file and a microphone to see if it works?

dpopp07:
Cool. I ran tests with the file from the examples in resources but I haven’t tried with a microphone yet. I can try that this morning

germanattanasio:
@dpopp07 @anweshan where are we with this?

dpopp07:
The current status is that all uses cases should be working, with the exception of users that are returning request objects in order to pipe them into a writable stream.

PR #730, opened 2018-06-26
update stt example

dpopp07:
The current example for STT is out of date. It includes a couple deprecated events, which I removed. I also added some comments explaining the two different ways to use the stream:
1. Getting the full response back as an object
2. Getting just the transcriptions as strings

The last thing I did was change the testing console statements to be what I felt was cleaner and would allow for printing stringified objects.

### DIFF
--- a/examples/speech_to_text.v1.js
+++ b/examples/speech_to_text.v1.js
@@ -1,5 +1,4 @@
 'use strict';
-
 var SpeechToTextV1 = require('watson-developer-cloud/speech-to-text/v1');
 var fs = require('fs');
 
@@ -9,8 +8,17 @@ var speechToText = new SpeechToTextV1({
   url: 'https://stream.watsonplatform.net/speech-to-text/api/'
 });
 
+/*
+    This code will print the entire response to the console when it
+    receives the 'data' event. Some applications will want to write
+    out only the transcribed text, to the console or to a file.
+    To this, remove `objectMode: true` from the `params` object.
+    Then, uncomment the block of code at Line 30.
+*/
+
 var params = {
-  content_type: 'audio/wav'
+  content_type: 'audio/wav',
+  objectMode: false
 };
 
 // create the stream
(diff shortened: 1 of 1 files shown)


examples/speech_to_text.v1.js
```diff
@@ -9,8 +8,17 @@ var speechToText = new SpeechToTextV1({
   url: 'https://stream.watsonplatform.net/speech-to-text/api/'
 });
 
+/*
+    This code will print the entire response to the console when it
+    receives the 'data' event. Some applications will want to write
+    out only the transcribed text, to the console or to a file.
+    To this, remove `objectMode: true` from the `params` object.
```
jeffpk62: "To do this" (just missed 'do').

examples/speech_to_text.v1.js
```diff
@@ -19,19 +27,20 @@ var recognizeStream = speechToText.createRecognizeStream(params);
...
 
-// and pipe out the transcription
+/*
+// pipe out the transcription to a file
 recognizeStream.pipe(fs.createWriteStream('transcription.txt'));
 
-// listen for 'data' events for just the final text
-// listen for 'results' events to get the raw JSON with interim results, timings, etc.
+// get strings instead of Buffers from `data` events
+// only do this when objectMode is `false`
```
jeffpk62: I'd take this line of the comment and move it to line 30.  You could then intro both lines of code with some reference to everything that's commented.  I think you want to call both `pipe` and `setEncoding` or neither of the two, right?

examples/speech_to_text.v1.js
```diff
@@ -19,19 +27,20 @@ var recognizeStream = speechToText.createRecognizeStream(params);
...
+recognizeStream.setEncoding('utf8');
+*/
 
-recognizeStream.setEncoding('utf8'); // to get strings instead of Buffers from `data` events
+recognizeStream.on('data', function(event) { onEvent('Data:', event); });
```
jeffpk62: I like this approach, Dustin.  At a minimum, I think it's cleaner and easier to follow at first glance.  Thanks!

examples/speech_to_text.v1.js
dpopp07: Good catch, thanks!

examples/speech_to_text.v1.js
dpopp07: Thanks - I copied that from your code haha

examples/speech_to_text.v1.js
dpopp07: That's a good idea, because they both apply to object mode being `false`. To clarify though, they do not have to be used together - you can set the encoding to log strings to the console without using the `pipe` statement to send it to a file. On the flip side, you can write the stream to a file without setting the encoding, but I imagine the user wouldn't find the results too helpful 🙂

jeffpk62: Thanks for the extra details, Dustin!

examples/speech_to_text.v1.js
jeffpk62: You know, I thought it looked familiar... :-)

PR #734, opened 2018-06-28
support access_token query param in recognize-stream

dpopp07:
Addressing the issue about using `access_token` instead of `watson-token` when using IAM.

This change feels a little weird because I don't see a large use case for it. We require a form of authentication when the user instantiates the service, and that works to authenticate the WS connection with the `Authorization` header. That being said, now if the user provides an `access_token` as a param, it will be used and the request for the token to put in the header will not be made.

This works to authenticate the request, but I am getting this warning from the service:
```
  "warnings": [
    "Unknown URL query params: access_token. Websockets requests should have the parameters in WS messages, not in URL."
  ]
```
which I thought was odd.

Would appreciate feedback on the implementation. I will add tests also, but wanted to make sure that this is what we want.

### DIFF
--- a/lib/querystring.ts
+++ b/lib/querystring.ts
@@ -14,7 +14,7 @@ const stringify = (queryParams: Object): string => {
       return (
         key +
         '=' +
-        (key === 'watson-token'
+        (key === 'watson-token' || key === 'access_token'
           ? queryParams[key]
           : encodeURIComponent(queryParams[key]))
       ); // the server chokes if the token is correctly url-encoded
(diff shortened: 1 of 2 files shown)


jeffpk62:
@dpopp07 See my email about this exact issue.  Since the Node SDK is object-based and requires an instance of the SpeechToTextV1 object to call `recognizeStream()` at all, I'm not sure of this parameter's utility?  I would definitely hold off on the merge until we get further guidance.

dpopp07:
I agree, the instantiation of the service alone already requires authentication. I noted this concern in my original PR message, as you can see. I think holding off for now is a good idea! Would like to get another opinion on this

jeffpk62:
Yep, I saw your comment after I'd written essentially the same in email, @dustinpopp. :-)  Definitely, we need another opinion.  Thank you!

dpopp07:
After discussing with Taj and German, it is decided that the current authorization implementation is sufficient.

PR #735, opened 2018-06-28
Add deprecation notices for Language Translator v2

lpatino10:
This PR adds deprecation notices for Language Translator v2 in the README and service class.

This will also be the final change for release 3.5.2.

### DIFF
--- a/package.json
+++ b/package.json
@@ -121,13 +121,13 @@
     "**/*.ts": [
       "tsc",
       "tslint --fix -p . -c tslint.json",
-      "test-unit",
+      "npm run test-unit",
       "git add"
     ],
     "test/**/*.js": [
       "eslint --fix",
       "prettier --write",
-      "test-unit",
+      "npm run test-unit",
       "git add"
     ],
     "examples/*.js": [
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -121,13 +121,13 @@
     "**/*.ts": [
       "tsc",
       "tslint --fix -p . -c tslint.json",
-      "test-unit",
+      "npm run test-unit",
```
dpopp07: What are these changes for? Did they fix the commit-hook for you?

lpatino10: Looks like I accidentally went up a major version of `lint-staged`. I got this error:
```
`lint-staged` no longer supports running scripts defined in package.json.

The same behavior can be achieved by changing the command to any of the following:
  - `npm run test-unit -- `
  - `mocha test/unit/ `
```
corresponding with the breaking change here: https://github.com/okonet/lint-staged/releases/tag/v7.0.0

I can go back to the old version if we'd rather not have that.

lpatino10: Although it looks like we were already at 7.2.0 actually

dpopp07: Okay, that's fine then. Just wanted to clarify. I haven't gotten to that stage of the commit-hook in a while haha

PR #736, opened 2018-06-28
Add semantic-release integration

lpatino10:
This PR adds `semantic-release` integration to automate SDK releases. With this, the SDK will be released with successful `master` builds (assuming `semantic-release` determines there should be a release - see [here](https://github.com/semantic-release/commit-analyzer#release-rules)).

Also added in this PR is `commitlint` to ensure commit messages follow the proper format, since it's necessary for `semantic-release` to work properly. Here's an example of `commitlint` in action:

bad_commit

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,3 +21,5 @@ deploy:
 - provider: script
   skip_cleanup: true
   script: npx travis-deploy-once "npx semantic-release"
+  on:
+    node: stable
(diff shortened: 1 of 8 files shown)


.travis.yml
```diff
@@ -8,6 +8,7 @@ before_install:
 - '[ "${TRAVIS_PULL_REQUEST}" = "false" ] && openssl aes-256-cbc -K $encrypted_320e4a7e27b3_key
   -iv $encrypted_320e4a7e27b3_iv -in auth.js.enc -out test/resources/auth.js -d ||
   true'
+- npm install npm@latest -g
```
dpopp07: Not sure about this line. Travis should already have `npm` installed and installing the "latest" every time had the potential to introduce breaking changes if they do a major release (unlikely, but still possible).

Did you need to add this to get it to work?

PR #743, opened 2018-07-12
docs: add deprecation message for using watson-tokens

dpopp07:
Added the deprecation message in the README, and within the block that checks if the user passed `token` or `watson-token` as a query param.


mediumTaj:
Can you take a look at why these tests are failing?

dpopp07:
Tests passed!

PR #746, opened 2018-07-16
4831/recognize stream comments

dpopp07:
@jeffpk62 noticed that the documentation for a few parameters in `RecognizeStream` was incorrect and did not match the behavior of the service - specifically, they were listing alternate default values for use of the service in object mode when the defaults are not actually different.

This PR addresses this inconsistency in the JSDoc comments.


dpopp07:
Agreed. I'll change the javascript sdk for now, but I will be looking into bringing it in as a dependency moving forward. Or at the very least, changing the javascript sdk to rely more on this sdk so that there is less repeated code.

PR #749, opened 2018-07-19
refactor: change format of token manager export

dpopp07:
Re-formatted the token manager to look more like the generated SDKs. This allows for a simpler import of the token manager as a module:
```js
const IamTokenManagerV1 = require('../../iam-token-manager/v1');
  // instead of
const IamTokenManagerV1 = require('../../iam-token-manager/v1').IamTokenManagerV1;
```

Tests and base service adjusted to reflect this.

### DIFF
--- a/iam-token-manager/v1.ts
+++ b/iam-token-manager/v1.ts
@@ -17,28 +17,11 @@
 import extend = require('extend');
 import { sendRequest } from '../lib/requestwrapper';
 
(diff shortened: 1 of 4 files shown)


dpopp07:
Closing to avoid breaking change and to adhere to to typescript good practice
https://basarat.gitbooks.io/typescript/docs/tips/defaultIsBad.html

PR #757, opened 2018-07-26
test: add unit tests for recognize() and createRecognizeStream()

dpopp07:
Add unit tests for `recognize()` and `createRecognizeStream()`. Also commented out a set of broken tests that were causing mocha to hang. These can be renovated later but for now should not be run. Also added some extra clarifying comments.

### DIFF
--- a/test/unit/test.speech_to_text.v1.js
+++ b/test/unit/test.speech_to_text.v1.js
@@ -1,10 +1,12 @@
 'use strict';
 
 const assert = require('assert');
-const watson = require('../../index');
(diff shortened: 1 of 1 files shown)


test/unit/test.speech_to_text.v1.js
```diff
@@ -153,7 +181,7 @@ describe('speech_to_text', function() {
   });
 
   // this test is severely broken
-  describe('recognizeStream()', function() {
+  /* describe.skip('recognizeStream()', function() {
```
germanattanasio: I think you can remove this test.

PR #761, opened 2018-08-01
Update mic stream example

dpopp07:
During a support session, I discovered that the example for using STT to stream from the microphone to the console was broken. This was due to the `interim_results` parameter not being passed in as `true`, as it defaults to `false`.

I updated the example to use the correct parameter and changed the JSDocs for `RecognizeStream` because they incorrectly listed the default value for `interim_results` as `true`.

ref:

From the [Speech to Text](https://www.ibm.com/watson/developercloud/speech-to-text/api/v1/curl.html?curl#websocket_methods) docs:
> **interim_results**
If `true`, the service returns interim results as a stream of JSON `SpeechRecognitionResults` objects. If `false`, the service returns a single `SpeechRecognitionResults` object with final results only.
Default: `false`

cc @jeffpk62

### DIFF
--- a/examples/browserify/package-lock.json
+++ b/examples/browserify/package-lock.json
@@ -1,0 +1,3416 @@
+{
+  "name": "watson-developer-cloud-browserify-example",
+  "version": "1.0.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@types/caseless": {
+      "version": "0.12.1",
+      "resolved": "https://registry.npmjs.org/@types/caseless/-/caseless-0.12.1.tgz",
+      "integrity": "sha512-FhlMa34NHp9K5MY1Uz8yb+ZvuX0pnvn3jScRSNAb75KHGB8d3rEU6hqMs3Z2vjuytcMfRg6c5CHMc3wtYyD2/A=="
+    },
+    "@types/csv-stringify": {
+      "version": "1.4.2",
+      "resolved": "https://registry.npmjs.org/@types/csv-stringify/-/csv-stringify-1.4.2.tgz",
+      "integrity": "sha512-OxBpngVW09fD5S90Fi2ihWJa9gWk/cYSeZ05T5PiElOJw4OjKq9bXyVEkEPpHS/1Vr/gKzAgv7okuvHLpb5CSA==",
+      "requires": {
+        "@types/node": "*"
+      }
+    },
+    "@types/extend": {
+      "version": "3.0.0",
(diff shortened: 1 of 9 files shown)


dpopp07:
You mean the `package.json` for the `examples/` folder?

It has the line `"watson-developer-cloud": "*",` which should always be pulling in the latest version.

anweshan:
@dpopp07 No, I mean the package.json in the `examples/speech_to_text_microphone_input` folder which should be changed to what you said above

anweshan:
Actually I just noticed that all the package.json files inside the examples folder inside the examples that have their own package.json inside their individual folder should probably be updated with that

dpopp07:
Also, I verified that the examples work with the updated SDK versions.

jeffpk62:
@dpopp07 Sorry for not seeing that comment about `interim_results` sooner.  I'd have pointed it out otherwise.  The API reference has the correct default (`false`).

dpopp07:
@jeffpk62 No worries! I just tagged you so you would be aware of the fix.

All - holding off on merge until #762 is merged, since that should prevent the build on `master` from failing.

qunliu:
Is there a similar issue with another sample STT code? (https://github.com/watson-developer-cloud/node-sdk/blob/master/examples/speech_to_text_microphone_input/transcribe-mic-to-file.js)

I pulled down the sample code and ran it, the recorded mic wav file sounds funny (like a slow mo recording where I record 2 seconds, but it gives me 8 seconds or something). However, I put interim_results there doesn't help. Did I miss anything else? Would you please take a look? Thanks!

dpopp07:
Hi @qunliu, I tested that example and found that it was not affected by the above problem. This example is for listening to audio from the microphone and writing the transcribed words to a text file. You should not be expecting an audio file.

If you are having trouble with a text-to-speech example, please open an issue with your code and I will be happy to take a look!

qunliu:
Thanks for looking into it @dpopp07 I am a still bit confused. If I understand this example correctly, it does a few things (please see the screenshot below):
* Listen to a mic and get audio stream from it;
* Pipe this audio stream to a file writer and get an audio file written to disk;
* Pipe the audio stream once again to Watson speech to text service and once the transcription is back, it gets written to a file (note: I am not looking at text to speech).

In my test, the transcription was all wrong, so that's where I started looking at that audio file recorded on disk and noticed that was wrong as well which in turn gave wrong transcription.

Sorry if I totally misunderstood the example. Please let me know what you think.

Thanks!

image

dpopp07:
Hi @qunliu, I now see what you are saying. I misunderstood your problem at first. Thank you for clarifying. I opened an issue to track this while I investigate - Please reference #765

PR #762, opened 2018-08-02
add semantic-release and @semantic-release/changelog

dpopp07:
The build of `master` is failing due to a lack of `@semantic-release/changelog` as a module. I thought this would be included by default but [turns out, it is not](https://github.com/semantic-release/semantic-release/blob/caribou/docs/extending/plugins-list.md).

Adding this to fix.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -388,6 +388,49 @@
         "rimraf": "^2.5.2"
       }
     },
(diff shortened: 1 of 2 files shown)


dpopp07:
No worries! There was nothing in the `changelog` repo that even mentioned needing to install so it was an easy miss. Thanks for the review!

PR #771, opened 2018-08-23
Node 4 compatibility - DO NOT MERGE

dpopp07:
I'm not sure why isn't hasn't broken the build until now, but `semantic-release` is incompatible with version 4 of Node. This is breaking the Travis build when it is testing Node 4. I tried a number of different options for getting around this, and I settled on this one as the cleanest and most efficient.

We don't need `semantic-release` or its associated plugins in Node 4 - only in Node 8 when we deploy. This PR makes `semantic-release` an "optional" dependency. When we run the install script for Node 4 (based on the new condition in the `.travis.yml` file), we run it with the `--no-optional` flag so that it doesn't install them. Otherwise, it will.

My only concern is that when users install `watson-developer-cloud` with `npm`, it will try to install these and break them. I'm not sure though.

Feedback is welcome. This is a kind of tricky and annoying problem.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,7 @@ before_install:
   true'
 - npm install -g typescript
 install:
-- echo $TRAVIS_NODE_VERSION
-- 'if [ "${TRAVIS_NODE_VERSION}" = "4.9.1" ]; then npm install --no-optional; else npm install; fi;'
+- 'if [ "${TRAVIS_NODE_VERSION}" = "4" ]; then npm install --no-optional; else npm install; fi;'
 script:
 - tsc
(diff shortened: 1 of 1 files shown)


germanattanasio:
I think you can remove semantic release from the dependencies and just leave the 
```
  script: npx semantic-release
```
in Travis. `npx` will install the dependency if it doesn't exist

dpopp07:
I thought about that but I don't think it will install the `@semantic-release/changelog` or `@semantic-release/git` plugins. I will play around with that to see if there's a way to specify those too. Good idea

dpopp07:
This may not be necessary, the build is passing in `master`. We'll see next time another PR is opened.

PR #776, opened 2018-09-04
Application security on cloud

mediumTaj:
This pull request enables AppScan on tagged releases.


germanattanasio:
is this the code to run the code-scan? can we move it to a library? that way we can use it in other sdks

dpopp07:
Yes, that's an open issue that we haven't found time for yet. That's why I haven't merged this

npacucci:
So this will fix also [this](https://github.com/watson-developer-cloud/node-sdk/issues/884) for Assistant2?

PR #778, opened 2018-09-11
Regenerate sdk release 7

dpopp07:
Regenerate with latest changes to API definitions and Watson Swagger Codegen.

### DIFF
--- a/assistant/v1.ts
+++ b/assistant/v1.ts
@@ -290,7 +290,7 @@ class AssistantV1 extends BaseService {
    * @param {number} [params.page_limit] - The number of records to return in each page of results.
    * @param {boolean} [params.include_count] - Whether to include information about the number of records returned.
    * @param {string} [params.sort] - The attribute by which returned results will be sorted. To reverse the sort order,
-   * prefix the value with a minus sign (`-`). Supported values are `name`, `updated`, and `workspace_id`.
+   * prefix the value with a minus sign (`-`).
    * @param {string} [params.cursor] - A token identifying the page of results to retrieve.
    * @param {boolean} [params.include_audit] - Whether to include the audit properties (`created` and `updated`
    * timestamps) in the response.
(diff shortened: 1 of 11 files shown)


assistant/v1.ts
```diff
@@ -2332,7 +2332,7 @@ class AssistantV1 extends BaseService {
    * contain carriage return, newline, or tab characters, and it must be no longer than 2048 characters.
    * @param {string} [params.new_parent] - The ID of the parent dialog node.
    * @param {string} [params.new_previous_sibling] - The ID of the previous sibling dialog node.
-   * @param {Object} [params.new_output] - The output of the dialog node. For more information about how to
+   * @param {DialogNodeOutput} [params.new_output] - The output of the dialog node. For more information about how to
```
mediumTaj: NICE!

germanattanasio:
I think you forgot to apply the manual changes for the create classifier in VR. The models are wrong

dpopp07:
@germanattanasio I applied [the manual changes](https://github.ibm.com/MIL/watson-swagger-codegen/wiki/NodeJS-Post-Generation-Notes#visual-recognition), which are just to parse out examples with custom names. What is new is the filename parameters, which the generator now adds to every file type parameter in the SDK.

I can see how this may be problematic, so either I can remove the filename code or edit it so that the code parses custom filenames the same way it's parsing for the examples. Both would require additional manual changes. Let me know what you think

cc @mkistler

visual-recognition/v3-generated.ts
```diff
@@ -414,6 +432,8 @@ class VisualRecognitionV3 extends BaseService {
    * depict the visual subject of any of the classes of the new classifier. Must contain a minimum of 10 images.
    *
    * Encode special characters in the file name in UTF-8.
+   * @param {string} [params.classname_positive_examples_filename] - The filename for classname_positive_examples.
```
germanattanasio: see this description

visual-recognition/v3-generated.ts
```diff
@@ -228,6 +232,8 @@ class VisualRecognitionV3 extends BaseService {
...
```
germanattanasio: see this descriptions

anweshan:
Were the warnings in Visual Recognition generated?

dpopp07:
The warnings are generated when `x-include-filename` is `true` in the Swagger, which means the filename is technically required but we didn't want to introduce breaking changes - hence the warning.

The _only_ change in VR is adding the filename params. There are no differences in manual changes. German is suggesting updates to the documentation for the filename params. What do you think won't work?

dpopp07:
@germanattanasio I realized that the filename parameter wouldn't match the positive example parameter name since those are custom. I made some edits to resolve this and would like your opinion. I think this works well but it will make it harder to generate this code in the future. Let me know what you think

PR #779, opened 2018-09-20
Regenerate sdk release 8

dpopp07:
- Regenerate applicable services
- Add Assistant V2 + tests
- Add option to disable SSL

Could use validation against ICP


anweshan:
@dpopp07 Does  there need to be something in the README about ssl option? Since the code is in base-service it isn't as clear

mediumTaj:
Is this ok to release?

dpopp07:
Yes

PR #782, opened 2018-09-27
Disable ssl for websocket

dpopp07:
The SSL disabling option I added for the last release did not cover the WebSocket connection (thanks @lpatino10 for the catch).

This PR resolves this by taking the user option, passing it to the `RecognizeStream` class, and using it to configure the connection. It does this by creating a `tlsOptions` parameter, which gets passed into Node's native `http` library [according to the `websocket` docs](https://github.com/theturtle32/WebSocket-Node/blob/master/docs/WebSocketClient.md#client-config-options).

@lpatino10 confirmed this fix works with ICP.

cc @germanattanasio

### DIFF
--- a/lib/recognize-stream.ts
+++ b/lib/recognize-stream.ts
@@ -226,17 +226,13 @@ class RecognizeStream extends Duplex {
     // for the last argument, `tlsOptions` gets passed to Node's `http` library,
     // which allows us to pass a rejectUnauthorized option
     // for disabling SSL verification (for ICP)
-    const clientConfig = options.rejectUnauthorized
-      ? { tlsOptions: { rejectUnauthorized: false }}
-      : null;
-
     const socket = (this.socket = new w3cWebSocket(
       url,
       null,
       null,
       options.headers,
       null,
-      clientConfig      
+      { tlsOptions: { rejectUnauthorized: options.rejectUnauthorized }}
     ));
 
     // when the input stops, let the service know that we're done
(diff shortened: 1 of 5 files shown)


lib/recognize-stream.ts
```diff
@@ -117,6 +117,7 @@ class RecognizeStream extends Duplex {
    * @param {string} [options.base_model_version] - The version of the specified base model that is to be used with recognition request or, for the **Create a session** method, with the new session.
    * Multiple versions of a base model can exist when a model is updated for internal improvements. The parameter is intended primarily for use with custom models that have been upgraded for a new base model.
    * The default value depends on whether the parameter is used with or without a custom model. For more information, see [Base model version](https://console.bluemix.net/docs/services/speech-to-text/input.html#version).
+   * @param {Boolean} [options.rejectUnauthorized] - If true, disable SSL verification for the WebSocket connection
```
germanattanasio: is `rejectUnauthorized` the name we use in the other SDKs?

lib/recognize-stream.ts
```diff
@@ -217,14 +220,19 @@ class RecognizeStream extends Duplex {
...
+    // which allows us to pass a rejectUnauthorized option
+    // for disabling SSL verification (for ICP)
     const socket = (this.socket = new w3cWebSocket(
       url,
       null,
       null,
       options.headers,
-      null
+      null,
+      options.rejectUnauthorized ? { tlsOptions: { rejectUnauthorized: false }} : null
```
lpatino10: Looks like the ternary expressions need to be swapped. Right now we're setting `rejectUnauthorized` to false if it's true.

dpopp07: Oh shoot I did something silly. I'm not supposed to be setting `rejectUnauthorized` at all. I'll fix that now.

lib/recognize-stream.ts
dpopp07: Changed 👍

PR #788, opened 2018-10-08
Regenerate sdk release 9

dpopp07:
- Regenerate with new changes to service
- Add deprecation messages for Conversation V1, `api_key` in VisRec, and LTv2 (there was JSDoc for this but never a console message).


visual-recognition/v3-generated.ts
```diff
@@ -611,7 +610,6 @@ namespace VisualRecognitionV3 {
   export type Options = {
     version: string;
     url?: string;
-    api_key?: string;
```
mediumTaj: Do we need to keep this to maintain compatibility until next major release?

dpopp07: We shouldn't need to since it is available in the base class options but I will confirm

dpopp07: Confirmed, this still transpiles in TypeScript.

PR #792, opened 2018-10-12
fix(discovery): `collection_ids` is now kept as an array DONT MERGE

dpopp07:
Thanks to @lpatino10 for the catch - `collection_ids` is meant to be an array. This patch stops it from being converted to a string.

Test added to verify.


dpopp07:
Fixed by the service - no longer needed

PR #795, opened 2018-10-23
Remove compatibility layer

dpopp07:
Removes the compatibility layer for every service that had one. Updates tests and examples to reflect these changes. In most cases, I deleted the old `v1.ts` and renamed `v1-generated.ts` to `v1.ts` so the user experience stays the same.

For the utility methods in STT and TTS, I decided to leave the import structure how it is for ease of development. I may look into other methods in the future. This way, if we ever need similar compatibility layers (which I don't love, but are actually a pretty good way to handle hand-written changes) in the future, we still have this pattern.


germanattanasio:
@dpopp07 I'm ok merging this if you show me the migration guide. I want to have that as part of this so that I can validate we are documenting all the breaking changes

dpopp07:
I am noting which services have breaking changes in the commit messages so that they will appear in the changelog but I agree, a detailed migration guide would be helpful. I will start working on that

PR #806, opened 2018-11-20
Patch stt and discovery

dpopp07:
Updating STT and Discovery parameters based on recent changes to the service


dpopp07:
fyi @jeffpk62

PR #807, opened 2018-11-29
feat(text-to-speech): add support for using `synthesize`

dpopp07:
Adds a new method - `synthesizeUsingWebSocket` - that returns a `ReadableStream` that the user can pipe into a new file or over the speakers. Returning a stream seemed to be the best user experience since the point of websocket is to allow the user to receive data in chunks in real time without reading all data into memory at once.

I tried to keep the design similar to `recognizeUsingWebSocket` for a consistent experience, but did not adopt any code that seemed unnecessary.

There might be more features that are needed but this provides the basic needs of the user with simple code and a simple interface.

I still need to write tests and an example but wanted to start getting feedback on the implementation.

cc @ehdsouza @jsstylos @jeffpk62 

Todo:
- [x] Write tests
- [x] Write an example

---

This is what the experience looks like for the user:
```js
const tts = new TTS({
  username,
  password,
});

const params = {
  text: 'Hello world.',
  accept: 'audio/ogg;codecs=opus',
};

// create the stream
const synthStream = tts.synthesizeUsingWebSocket(params);

// pipe the stream into a new audio file
synthStream.pipe(fs.createWriteStream('./speech.ogg'));

// optionally, the user can listen to data as it is returned from the service
synthStream.on('message', (message, data) => {
  console.log(data);
});
```


germanattanasio:
How can you see when the websocket was closed? can we have an `on('close')`?

dpopp07:
@germanattanasio documentation, example, and test added

examples/text_to_speech_websocket.js
```diff
@@ -0,0 +1,51 @@
...
+// 'err' is the Error object describing the error
+synthesizeStream.on('error', err => {
+  console.log(err);
+});
+
+// the 'close' event is emitted once, when the connection is terminated by the service
+// the 'code' parameter is the status code. 1000 is the code for a normal termination
+// the 'reason' parameter provides a string description of how the connection closed
+synthesizeStream.on('close', (code, reason) => {
+  console.log(code);
```
germanattanasio: Nice 🎉

germanattanasio:
When running in node we use the `ws` package but do we create a bundle for the browser without it? and is the `ws` interface the same as the WebSocket interface in a browser?

The questions are not related to this PR which is good

PR #813, opened 2018-12-12
add support for promises

dpopp07:
Adds support for returning Promises in the SDK methods _in addition to callbacks_.

This is accomplished by returning a Promise anytime a callback in not defined. To be as unobtrusive as possible, the Promise argument is a function that calls the method again but _with_ a callback. The results passed to the callback are used to fulfill the Promise. This allows all of the rest of the code to run as it was before without modification.

Returning Promises (regardless of the implementation) comes with one major drawback - only one value can be resolved from the Promise. This is an issue because we typically provide both the response body and the entire response object to the user in the form of callback arguments. To be consistent, the user should have the option of accessing either when using Promises.

One solution is to return a wrapper object that contains both, but since the user will typically only want the body, it seemed like bad design to introduce an extra step of unwrapping the body with every API call. The solution I settled on is to introduce an optional parameter that the user can pass in if they want to receive the entire response object rather than just the body.

Looking for feedback on this design before I implement with the rest of the services.

Todo:
- [x] Add this logic to the rest of the services
- [x] Add unit tests for Promise returning logic
- [x] Add some more integration tests to ensure logic works
- [x] Add documentation


assistant/v2.ts
```diff
@@ -76,16 +76,24 @@ class AssistantV2 extends BaseService {
    * @param {Function} [callback] - The callback that handles the response.
    * @returns {NodeJS.ReadableStream|void}
    */
-  public createSession(params: AssistantV2.CreateSessionParams, callback?: AssistantV2.Callback<AssistantV2.SessionResponse>): NodeJS.ReadableStream | void {
+  public createSession(params: AssistantV2.CreateSessionParams, callback?: AssistantV2.Callback<AssistantV2.SessionResponse>): NodeJS.ReadableStream | Promise<any> | void {
     const _params = extend({}, params);
-    const _callback = (callback) ? callback : () => { /* noop */ };
+    const _callback = (callback) ? callback : undefined;
```
germanattanasio: you don't need this. you are basically setting it as undefined but you can just check of trueness

assistant/v2.ts
```diff
@@ -76,16 +76,24 @@ class AssistantV2 extends BaseService {
...
     const requiredParams = ['assistant_id'];
 
     const missingParams = getMissingParams(_params, requiredParams);
     if (missingParams) {
       return _callback(missingParams);
```
germanattanasio: Missing parameters won't get caught in a `reject()`.

germanattanasio: Check that if we have errors in the previous statements we always return a reject.

dpopp07: Great catch, I feel silly for missing this. All I have to do is move the Promise return function to before the missing parameter check and it will call the reject properly

assistant/v2.ts
```diff
@@ -76,11 +76,19 @@ class AssistantV2 extends BaseService {
...
+  public createSession(params: AssistantV2.CreateSessionParams, callback?: AssistantV2.Callback<AssistantV2.SessionResponse>): NodeJS.ReadableStream | Promise<any> | void {
     const _params = extend({}, params);
-    const _callback = (callback) ? callback : () => { /* noop */ };
+    const _callback = callback;
     const requiredParams = ['assistant_id'];
 
+    if (!_callback) {
+      return new Promise((resolve, reject) => {
+        this.createSession(params, (err, bod, res) => {
+          err ? reject(err) : _params.return_response ? resolve(res) : resolve(bod);
```
germanattanasio: I would prefer to always return `body` and `response` as part of the response.
We are currently moving all the other SDKs to always return the HTTP response and the parsed body. I can't remember the names we are using but @lpatino10 or @ehdsouza can probably help you.

dpopp07: The issue is that a Promise can only resolve _one_ value. Should I be wrapping the whole respond and the parsed body up into one object and return that? Or just the response and let the user parse the body?

dpopp07: I don't think I understand. You're saying to return a single object that contains the results and the headers?

e.g.
```json
{
  "result": {
    "<body>"
  },
  "headers": {
    "<response headers>"
  }
}
```

We currently also return the request information which I think is beneficial for debugging, etc.

Why not just return the entire response if users have to go through an additional step to get to the body anyway?

germanattanasio: The entire response depends on the library that we currently use (`requests`) my proposal is to define the same interface we use with other sdks.

```json
{
  "result": <body-as-json-or-buffer>,
  "headers": {
    "header1": "value1"
  }
}
```

dpopp07: That's a good point. I'll wait to finish this PR until I finish converting the SDK to use `axios` instead of `request`.

Would you be okay with also returning the request object if possible? I find it can be helpful for debugging to see what the request looked like when it was sent:
```json
{
  "result": "<body-as-json-or-buffer>",
  "headers": {
    "header1": "value1"
  },
  "request": " <request object>"
}
```

dpopp07: The default response structure for `axios` looks like this:
```
{
  // `data` is the response that was provided by the server
  data: {},

  // `status` is the HTTP status code from the server response
  status: 200,

  // `statusText` is the HTTP status message from the server response
  statusText: 'OK',

  // `headers` the headers that the server responded with
  // All header names are lower cased
  headers: {},

  // `config` is the config that was provided to `axios` for the request
  config: {},

  // `request` is the request that generated this response
  // It is the last ClientRequest instance in node.js (in redirects)
  // and an XMLHttpRequest instance the browser
  request: {}
}
```

We might want to keep all of that information and rename `data` to `result` for consistency

germanattanasio: as long as we can change the library without breaking the interface

PR #814, opened 2018-12-13
chore: throw error if user makes common mistake

dpopp07:
It's been noticed that users will occasionally copy code from the documentation and assume that the characters `{` or `"` might belong at the beginning of their credentials.

Currently, when this happens, the request goes through and the service returns a 401 since the credentials are wrong. This PR adds more graceful handling to catch the problem before the request is sent, providing the user with a message about what might be wrong.

Example error messages with the current implementation:

Single credential bad parameter:
```
username starts with a bad character. Revise these credentials - they should not start with a { or "
```

Multiple bad parameters:
```
url starts with a bad character. username starts with a bad character. password starts with a bad character. Revise your credentials - they should not start with a { or "
```


jsstylos:
Maybe something more perscriptive like, "Credentials shouldn't start or end with curly brackets or quotes. Be sure to remove any { and \" characters surrounding your credentials". Or maybe @SirSpidey can come up with better wording.

Looking at the error, I don't know if we need to check the last character also, just so that users don't change only the first character and leave the quote or braces at the end.

dpopp07:
Those are good thoughts, was hoping to get feedback on the wording.

I'm good with checking the last characters also, I was just following the issue. Any other thoughts on this?

germanattanasio:
@dpopp07 Can you include the right way of calling the constructor? if you know there are `"` then remove them and show them in the error how to do it.

lib/base_service.ts
```diff
@@ -89,6 +89,30 @@ function usesBasicForIam(obj: any): boolean {
   return obj.username === 'apikey' && !obj.password.startsWith('icp-');
 }
 
+// returns true if the string has a { or " as the first character
+// these are common user-issues that we should handle before they get a network error
+function badFirstChar(value: string): boolean {
+  return value.startsWith('{') || value.startsWith('"');
```
mediumTaj: Can you check for `}` and `"` as the last character as well here? and change the name to reflect that

lib/base_service.ts
```diff
@@ -89,6 +89,30 @@ function usesBasicForIam(obj: any): boolean {
...
+}
+
+// checks credentials for common user mistakes of copying { or " characters from the documentation
+function checkCredentials(obj: any) {
+  let errorMessage = '';
+  const credsToCheck = ['url', 'username', 'password', 'iam_apikey'];
+  credsToCheck.forEach(cred => {
+    if (obj[cred] && badFirstChar(obj[cred])) {
+      errorMessage += `${cred} starts with a bad character. `;
```
mediumTaj: Can you change this to account for first and last characters?

lib/base_service.ts
```diff
@@ -89,6 +89,30 @@ function usesBasicForIam(obj: any): boolean {
...
+    }
+  });
+
+  if (errorMessage.length) {
+    errorMessage += 'Revise these credentials - they should not start with a { or "';
```
mediumTaj: Can you change this to account for first and last characters?

lib/base_service.ts
mediumTaj: >The ${cred} shouldn't start or end with curly brackets or quotes. Be sure to remove any {} and \"

PR #819, opened 2019-01-07
test: remove pi v2 test and example - service is no longer available

dpopp07:
It seems that the Node SDK missed the memo when the other SDKs dropped Personality Insights v2 because it still supports the service (the other SDKs do not).

This version of the service is no longer available (returns 404 codes) so there is no reason to support it, test it or provide an example.

I thought it might be better to drop the actual code with the next major release, but the test and example should be dropped now. If we should just go ahead and the drop the code with this PR, I'm happy to make that edit.


dpopp07:
Good call, was using them locally but had not yet updated Travis

PR #829, opened 2019-01-23
refactor core code to use `axios` instead of `request`

dpopp07:
This PR migrates the Node SDK to use [`axios`](https://github.com/axios/axios) instead of [`request`](https://github.com/request/request) for making network requests. This involved some minor modifications to the code:
- Constructing a config object formatted for `axios` with the request data
- Building FormData parameters internally, since this is not handled by `axios`
- Adding `responseType` to several SDK methods that return Streams (this should be generated in the future)
- Refactoring tests that relied on `request` returning a Stream, as `axios` does not return streams
- I also made slight code changes to improve readability

Looking for feedback now but there is still work to do.

What is still left to do?
- [x] Add code to perform the GZIPing that `request` did
- [x] Go through and make sure `responseType` is set everywhere it needs to be
- [x] Add code for setting `rejectUnauthorized` to `false` for ICP compatibility
- [x] Test with ICP
- [x] Determine an interface for the response
- [x] Determine an interface / format for errors as the error formatting method is currently not being used in the new code
- [x] Document any breaking changes in the UPGRADE file


lib/requestwrapper.ts
```diff
@@ -74,14 +76,14 @@ export function formatErrorIfExists(cb: Function): request.RequestCallback {
 
     // for api-key services
     if (response.statusMessage === 'invalid-api-key') {
-      const error = {
+      const err = {
         error: response.statusMessage,
         code: response.statusMessage === 'invalid-api-key' ? 401 : 400,
```
germanattanasio: this was needed for alchemy we can probably remove it now

lib/requestwrapper.ts
```diff
@@ -179,59 +156,111 @@ export function sendRequest(parameters, _callback) {
...
+
+  if (formData) {
+    data = multipartForm;
+    // form-data generates headers that MUST be included or the request will fail
+    headers = extend(true, {}, headers, multipartForm.getHeaders());
   }
 
-  // Compression support
-  options.gzip = true;
+  const axiosObject = {
```
germanattanasio: I would call it `requestParams`. I don't think we need `axios` to be part of the name

package.json
```diff
@@ -80,10 +80,12 @@
     "@types/node": "~10.3.5",
     "@types/request": "~2.47.1",
     "async": "~2.6.1",
+    "axios": "^0.18.0",
```
germanattanasio: Can you remove `request`?

lib/requestwrapper.ts
```diff
@@ -41,93 +44,92 @@ function parsePath(path: string, params: Object): string {
...
+      error['transaction-id'] = axiosError.headers['transaction-id'];
     }
 
-    // ensure a more descriptive error message
-    if (error && (error.code === 401 || error.code === 403)) {
-      error.body = error.message;
-      error.message = 'Unauthorized: Access is denied due to invalid credentials.';
+    // other services use x-global-transaction-id
+    if (axiosError.headers['x-global-transaction-id']) {
+      error['x-global-transaction-id'] = axiosError.headers['x-global-transaction-id'];
```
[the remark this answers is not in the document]
dpopp07: Yep, I can do that. Didn't know if that would make it too crowded but I like giving them that info

dpopp07:
@germanattanasio Work is complete and tests are updated. PR is ready for review

PR #831, opened 2019-02-01
Load creds from file

dpopp07:
Support loading credentials from a file called `ibm-credentials.env`. This file takes precedence over environment variables and VCAP services, but not over credentials given in the constructor. Follows a similar implementation as Logan's in Java.

Since the structure of this credentials file is the same as a standard `.env` file for environment variables, it made sense to me to re-use the `getCredentialsFromEnvironment` function for reading the credentials into the internal `options` variable. I had to slightly modify it so that I could pass in an object for it to look at.

The code will first look at the environment variable given by the user (if it exists): if it's a file, it will use that file. If it's a path, it will try to read <path>/ibm-credentials.env. Then it will look in the system home directory, then in the working directory (the directory the user ran the code from - most likely the projects root directory but not guaranteed).

Test cases and documentation added.


lib/read-credentials-file.ts
```diff
@@ -0,0 +1,51 @@
...
+    filepathToUse = homeDir;
+  } else if (fileExistsAtPath(workingDir)) {
+    filepathToUse = workingDir;
+  } else {
+    // file does not exist anywhere, will not be used
+    return {};
+  }
+
+  const credsFile = fs.readFileSync(filepathToUse);
+
```
germanattanasio: Right now it's impossible for the user to have some sort of logging.
Maybe we could add the `debug` npm package. I know it's outside of this issue but I think it will help users if they need to debug the SDK

lpatino10: ☝️ pretty good point. Probably something I should look at adding in my SDKs as well.

README.md
```diff
@@ -85,14 +85,55 @@ Watson services are migrating to token-based Identity and Access Management (IAM
...
+
+```bash
+export IBM_CREDENTIALS_FILE="<path>"
+```
+
+where `<path>` is something like `/home/user/Downloads/<file_name>.env`. If you just provide a path to a directory, the SDK will look for a file called `ibm-credentials.env` in that directory.
+
+#### Manually
+
+The SDK also supports setting credentials manually in your code. You will either use IAM credentials or Basic Authentication (username/password) credentials.
```
mediumTaj: ugh now i need to change my readme files, hah

PR #834, opened 2019-02-08
refactor: stop using cookies in requests

dpopp07:
The primary objective of this PR to remove the code that adds a cookie jar to the request. These changes are made in `lib/base_service.ts` and are small.

I noticed that I made a slight TypeScript blunder when converting the SDK to use `axios` instead of `request` - I left in the `request` types. This PR also includes the fix for that, changing the lines

```js
import { RequestResponse } from 'request';
// ...
export type Callback<T> = (error: any, body?: T, response?: RequestResponse) => void;
```
to
```js
import { AxiosResponse } from 'axios';
// ...
export type Callback<T> = (error: any, body?: T, response?: AxiosResponse<T>) => void;
```


dpopp07:
Yes, that's what I'm using for `AxiosResponse` 👍

PR #835, opened 2019-02-12
chore: Add setServiceUrl method to base service class

mkistler:
This PR adds a new method to the base service class to set the service URL after the service was initialized. All the other SDKs provide a method or means to do this, so this is just bringing the Node SDK up to par with the other SDKs.

##### Checklist

- [x] `npm test` passes (modulo some known problems)
- [x] tests are included

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ npm install watson-developer-cloud
 
 ## Usage
 
-The [examples][examples] folder has basic and advanced examples. The examples within each service assume that you already have [service credentials](#getting-credentials). 
+The [examples][examples] folder has basic and advanced examples. The examples within each service assume that you already have [service credentials](#getting-credentials).
 
 Credentials are checked for in the following order:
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -187,6 +187,18 @@ var discovery = new DiscoveryV1({
   });
 ```
 
+### Setting the service URL
+
+You can set the service URL by calling the setServiceUrl() method.
+
+```javascript
```
SirSpidey: So this means an update to our API references, where we have the following: (?)
```javascript
var DiscoveryV1 = require('watson-developer-cloud/discovery/v1');

var discovery = new DiscoveryV1({
  version: '{version}',
  iam_apikey: '{apikey}',
  url: 'https://gateway-wdc.watsonplatform.net/discovery/api'
});
```

dpopp07:
I don't think the front matter necessarily needs to be updated. The method you posted above still works and will be the preferred way of setting the URL. We don't document every available base class  method in the API reference so unless this is important, I'm not sure it's worth putting in there.

If you do think it should be there, that's fine with me. Just let me know before you do, since this change hasn't been released yet.

README.md
mkistler: Yes ... we can now offer the alternate means of setting the service URL for Node.

PR #837, opened 2019-02-14
build: update dependencies to latest versions

dpopp07:
Went through and updated all dependencies. Most dependency major version updates did not affect our code. The exception is `husky` which made breaking changes, so I updated the githooks accordingly.

Additionally, "babylon" is no longer available as a parser for `prettier`, it was being converted to "babel", so I made that change.

The only package I did not update is `lint-staged`, as its newest major version requires a Node version of 8 or higher. We are supporting Node 6, so we cannot upgrade that package.


germanattanasio:
@dpopp07 did you run `npm audit fix` that usually update the underlying dependencies for those dependencies that you can't update like `lint-stage`

dpopp07:
@germanattanasio Yes, I ran`npm audit fix` and it was able to fix 100% of the detected vulnerabilities 👍

PR #852, opened 2019-03-27
fix: allow users to use `iam_apikey` when authenticating for icp

dpopp07:
This PR bumps the version of `ibm-cloud-sdk-core` which adds functionality for users to authenticate with ICP using `iam_apikey`. This is should be the final update for v3 of this SDK, aside from bug fixes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,7 +87,7 @@
     "dotenv": "^2.0.0",
     "extend": "~3.0.2",
     "file-type": "^7.7.1",
-    "ibm-cloud-sdk-core": "0.0.1",
+    "ibm-cloud-sdk-core": "0.0.3",
     "isstream": "~0.1.2",
     "mime-types": "~2.1.18",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -87,7 +87,7 @@
     "dotenv": "^2.0.0",
     "extend": "~3.0.2",
     "file-type": "^7.7.1",
-    "ibm-cloud-sdk-core": "0.0.1",
+    "ibm-cloud-sdk-core": "0.0.3",
```
germanattanasio: Any reason why this is not a `^0.0.3`?

dpopp07: I didn't do that on purpose but it actually is probably for the best since we don't want it to do a minor bump unintentionally. `v0.1.0` introduces breaking changes to the core.

germanattanasio: :+1

PR #856, opened 2019-04-09
Publish to ibm-watson

dpopp07:
Updates `watson-developer-cloud` to `ibm-watson` wherever it is relevant (pretty much everywhere that isn't the repo url).

The package has been officially moved to `ibm-watson` and `watson-developer-cloud` has been officially deprecated on NPM.


germanattanasio:
Are we showing a message to instruct users to move to a different repo?
Take a look at https://stackoverflow.com/questions/46942043/npm-install-add-custom-warning-message

dpopp07:
I marked the [old package as deprecated](https://www.npmjs.com/package/watson-developer-cloud) with the following message:

> DeprecationWarning: watson-developer-cloud moved to ibm-watson. To get updates, use the new package.

It should print that message when the user installs the package. It is not doing that for me every time, so I am going to open an issue against `npm`. If we want a more obvious message, I can use the method you linked above.

dpopp07:
The issue I was seeing with NPM was just a problem with my cache. The deprecation message I noted above should print every time a user tries to install the package.

Screen Shot 2019-04-11 at 9 40 40 AM

PR #863, opened 2019-04-18
Rename files

dpopp07:
Now that we've removed the compatibility layers, we can remove the unnecessary placeholders and just use `vX.ts` for the filenames. The exception is STT and TTS, which have hand-written helper methods and therefore need to keep the old pattern.


dpopp07:
👍 I don't think we should, those helper functions are pretty necessary at the moment so they will stay as is for a while probably

PR #868, opened 2019-04-24
test: fix skipping two integration tests without auth.js

MasterOdin:
This fixes running the test suite locally when you have not created a test/resources/auth.js script as these two tests (integration/assistant.v2.test.js and integration/compare_comply.test.js) both relied on undefined elements of `auth` before we hit the describe skip blocks.

##### Checklist

### DIFF
--- a/test/integration/compare_comply.test.js
+++ b/test/integration/compare_comply.test.js
@@ -6,262 +6,283 @@ const authHelper = require('../resources/auth_helper.js');
 const auth = authHelper.auth;
 const describe = authHelper.describe; // this runs describe.skip if there is no auth.js file :)
 
-const compare_comply = auth ? new CompareComply(auth.compare_comply) : null;
-
-describe('html conversion', () => {
-  test('convertToHtml', done => {
-    const params = {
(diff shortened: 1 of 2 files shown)


test/integration/compare_comply.test.js
```diff
@@ -6,7 +6,7 @@ const authHelper = require('../resources/auth_helper.js');
 const auth = authHelper.auth;
 const describe = authHelper.describe; // this runs describe.skip if there is no auth.js file :)
 
-const compare_comply = new CompareComply(auth.compare_comply);
+const compare_comply = auth ? new CompareComply(auth.compare_comply) : null;
```
dpopp07: Instead of doing this, will you wrap all of the integration tests in a `describe('compare_comply_integration'...` block like we do in the other tests? Not sure why this one got left out originally but I would prefer all of the services to be consistent

MasterOdin: That makes sense and is definitely probably cleaner. Done.

dpopp07: Great, thanks!

PR #870, opened 2019-04-24
fix: added commitlint rules

### LINKED ISSUE #867: Include commitlint.config.js in repo
When attempting to make a commit to this repo, the commit message is checked by husky to ensure it follows proper paradigm. However, it's confusing that on the first attempt to make a commit, they get the following error message:
```
⧗   input: Test test test
✖   Please add rules to your `commitlint.config.js`
    - Getting started guide: https://git.io/fhHij
    - Example config: https://git.io/fhHip [empty-rules]
✖   found 1 problems, 0 warnings
    (Need help? -> https://github.com/conventional-changelog/commitlint#what-is-commitlint )
```
and now the user has to look at getting started for using commitlint and just hope that you folks aren't deviating from the default in any way.

joyychang:
- added `commitlint` rules. Resolves #867 
- updated some vulnerable dependencies per `npm audit` review.

##### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "lint-staged": "~7.2.0",
     "object.assign": "~4.1.0",
     "prettier": "^1.16.4",
-    "semantic-release": "^15.13.8",
+    "semantic-release": "^15.13.12",
     "tslint": "^5.12.1",
     "tslint-config-prettier": "^1.18.0",
     "tslint-eslint-rules": "^5.4.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -152,5 +152,72 @@
...
+        2,
+        "always",
+        72
+      ],
+      "scope-case": [
+        2,
+        "always",
+        "lower-case"
+      ],
+      "subject-case": [
```
dpopp07: What's your reasoning here? I feel like we want it to be always lower case

dpopp07: Realizing that this probably looks word by word, it makes sense to allow other cases to include variable names in commit messages

PR #873, opened 2019-04-30
test: fix running tests on non-ibm travis-ci

MasterOdin:
This fixes running the test suite on Travis-CI for forks as the first two arguments were throwing errors which in the case of the second line, caused the build to be left in the wrong folder causing the rest of the build to crash.

Examples:
Bad: https://travis-ci.com/MasterOdin/node-sdk/builds/110188118
Good: https://travis-ci.com/MasterOdin/node-sdk/builds/110188457

##### Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,10 +4,12 @@ node_js:
 - 6
 - 8
 before_install:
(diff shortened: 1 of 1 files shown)


germanattanasio:
This is awesome @MasterOdin Thanks for the PR

PR #881, opened 2019-05-09
Replace all Object types with more appropriate type

### LINKED ISSUE #880: Object is a wrong type for key-value pairs
**Overview**
I am using Node-SDK in my Typescript project and I discovered that Object type is used in definition of many fields containing key-value pairs.
According to https://stackoverflow.com/questions/18961203/any-vs-object
> * any can be anything (you can call any method etc on it without compilation errors)
> * Object exposes the functions and properties defined in the Object class.

Specifically I am trying to check property of [DialogNodeAction.parameters](https://github.com/watson-developer-cloud/node-sdk/blob/v4.1.0/assistant/v1.ts#L4356)

When I try to check if a property exists, I get error: TS2339: Property 'xxx' does not exist on type 'Object'.

I think that it would be better to replace  all Objects with `{ [index: string]: any }` in type definitions.

**Expected behavior**
It works

**Actual behavior**
S2339: Property 'xxx' does not exist on type 'Object'.

**SDK Version**
4.1.0

**Additional information:**
- OS: WSL on Windows 10
- v10.15.3

Naktibalda:
Closes #880

`Object` type exposes the functions and properties defined in the Object class only.
Changing signature to `{[key: string]: any}` permits to use arbitrary fields which, I guess, was original intention.
I used more specific `{[key: string]: string}` for `headers` field, but it was a guess, please let me know if it is wrong.

##### Checklist

### DIFF
--- a/assistant/v1.ts
+++ b/assistant/v1.ts
@@ -3293,7 +3293,9 @@ namespace AssistantV1 {
     /** The language of the workspace. */
     language?: string;
     /** Any metadata related to the workspace. */
-    metadata?: Object;
+    metadata?: {
+      [key: string]: any;
+    };
     /** Whether training data from the workspace (including artifacts such as intents and entities) can be used by IBM for general service improvements. `true` indicates that workspace training data is not to be used. */
     learning_opt_out?: boolean;
     /** Global settings for the workspace. */
@@ -3384,7 +3386,9 @@ namespace AssistantV1 {
     /** The language of the workspace. */
     language?: string;
     /** Any metadata related to the workspace. */
-    metadata?: Object;
+    metadata?: {
+      [key: string]: any;
+    };
     /** Whether training data from the workspace (including artifacts such as intents and entities) can be used by IBM for general service improvements. `true` indicates that workspace training data is not to be used. */
     learning_opt_out?: boolean;
     /** Global settings for the workspace. */
@@ -3683,7 +3687,9 @@ namespace AssistantV1 {
     /** The description of the entity. This string cannot contain carriage return, newline, or tab characters, and it must be no longer than 128 characters. */
     description?: string;
     /** Any metadata related to the entity. */
-    metadata?: Object;
+    metadata?: {
+      [key: string]: any;
+    };
     /** Whether to use fuzzy matching for the entity. */
     fuzzy_match?: boolean;
     /** An array of objects describing the entity values. */
@@ -3764,7 +3770,9 @@ namespace AssistantV1 {
     /** The description of the entity. This string cannot contain carriage return, newline, or tab characters, and it must be no longer than 128 characters. */
     new_description?: string;
     /** Any metadata related to the entity. */
-    new_metadata?: Object;
+    new_metadata?: {
+      [key: string]: any;
+    };
     /** Whether to use fuzzy matching for the entity. */
     new_fuzzy_match?: boolean;
     /** An array of objects describing the entity values. */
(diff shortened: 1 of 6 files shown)


germanattanasio:
@Naktibalda this looks good but we autogenerate this code so you will have to give us a few days to along with merging this code, update the generator.

@joyychang ☝️

dpopp07:
@germanattanasio Should we consider the use case of when a header is not a string? For example, `Content-Length`? I don't think that would get much usage but maybe we want to use slightly less specific `{[key: string]: any}` for headers too

dpopp07:
@mkistler said that Swift requires header values to be strings and that this is probably fine but if you have a different opinion German, let me know

germanattanasio:
I see your point with `Content-Length` will it be a problem if we set it to a string?

dpopp07:
Upon [further research],(https://stackoverflow.com/questions/34152142/possible-types-of-a-http-header-value) all headers are sent across HTTP as strings in the end. So, it is fine the way it stands.

I can make this change in the generator.

germanattanasio:
@dpopp07 did you make the change?

dpopp07:
I started working on it. Discovered a larger issue around this that needs to be fixed in the same PR. Should be going in next week

dpopp07:
Hey @Naktibalda sorry for the delay on this! We definitely want to get this in but I have a couple comments/suggestions that will improve auto-generation of this code. I would like to use named interfaces rather than defining the index types inline. I propose the following:

1. For the headers objects, using the predefined type of `OutgoingHttpHeaders` from `@types/node`. This gives us a well-defined type sanctified by Node. It is essentially the same type as in your change, with the addition of supporting number, string array, and undefined header values as well. [See definition.](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/http.d.ts#L67-L69)
```js
import { OutgoingHttpHeaders } from 'http';

// ...

export type Options = {
  version: string;
  url?: string;
  iam_access_token?: string;
  iam_apikey?: string;
  iam_url?: string;
  username?: string;
  password?: string;
  use_unauthenticated?: boolean;
  headers?: OutgoingHttpHeaders;
}
```

2. For the generic objects, I would like to define our own named interface `JsonObject`. This will hold the same type as in your change. The benefit is that it can easily be used everywhere it needs to, like JSDoc comments and interface declarations.

For example:
`* @param {Object} [params.metadata] - Any metadata related to the workspace.`
becomes
`* @param {JsonObject} [params.metadata] - Any metadata related to the workspace.`
instead of
`* @param {[key: string]: string} [params.metadata] - Any metadata related to the workspace.`

The interface would be defined with the other named interfaces present in each service:
```js
/** The callback for a service request. */
export type Callback<T> = (error: any, body?: T, response?: AxiosResponse<T>) => void;

/** The body of a service request that returns no response data. */
export interface Empty { }

/** A standard JS object, defined to avoid the limitations of `Object` and `object` */
export interface JsonObject {
  [key: string]: any;
}

// ...

/** Parameters for the `createWorkspace` operation. */
export interface CreateWorkspaceParams {
  /** The name of the workspace. This string cannot contain carriage return, newline, or tab characters, and it must be no longer than 64 characters. */
  name?: string;
  /** The description of the workspace. This string cannot contain carriage return, newline, or tab characters, and it must be no longer than 128 characters. */
  description?: string;
  /** The language of the workspace. */
  language?: string;
  /** Any metadata related to the workspace. */
  metadata?: JsonObject;
  /** Whether training data from the workspace (including artifacts such as intents and entities) can be used by IBM for general service improvements. `true` indicates that workspace training data is not to be used. */
  learning_opt_out?: boolean;
  /** Global settings for the workspace. */
  system_settings?: WorkspaceSystemSettings;
  /** An array of objects defining the intents for the workspace. */
  intents?: CreateIntent[];
  /** An array of objects describing the entities for the workspace. */
  entities?: CreateEntity[];
  /** An array of objects describing the dialog nodes in the workspace. */
  dialog_nodes?: DialogNode[];
  /** An array of objects defining input examples that have been marked as irrelevant input. */
  counterexamples?: Counterexample[];
  headers?: OutgoingHttpHeaders;
  return_response?: boolean;
}
```

I am open to suggestion on the implementation/naming here but this solution lends itself more properly to auto-generation. I created a branch called `alternate-object-type-example` with the [changes I am suggesting made to Assistant V1](https://github.com/watson-developer-cloud/node-sdk/blob/alternate-object-type-example/assistant/v1.ts), to be an example. Look at that code, try pulling it down and making sure it works for your purposes, and let me know what your thoughts are. Thanks again for the PR!

Naktibalda:
I haven't tested your code, but using OutgoingHttpHeaders and JsonObject type looks like good idea to me.

dpopp07:
@Naktibalda Great, we can move forward with that then. Do you want to update your PR to include these changes so we can merge it in? Or would you rather me open a separate PR?

Naktibalda:
Please make your own pull request.

PR #886, opened 2019-05-15
chore: remove detect-secret from repo

joyychang:


.travis.yml
```diff
@@ -4,18 +4,18 @@ node_js:
...
   -d || true'
-- '[ "${TRAVIS_PULL_REQUEST}" = "false" ] && [ ! -z ${encrypted_ac3aacad7ba8_key} ] 
+- '[ "${TRAVIS_PULL_REQUEST}" = "false" ] && [ ! -z ${encrypted_ac3aacad7ba8_key} ]
   && cd test/resources/ && tar xvf secrets.tar
   && cd ../.. || true'
 - npm install -g typescript
 script:
 - tsc
 - npm run test-unit-travis || travis_terminate 1
-- npm run test-integration-travis || node scripts/report_integration_test.js || (cat test-output.log && travis_terminate 1)
```
dpopp07: I'm good with this but wondering what your motives were for removing the output logging

joyychang: I added it previously cuz i was parsing stdout. now that i'm using json, the print is just ugly. Forgot to remove it with the previous PR

dpopp07:
@joyychang So when users don't have the files you are removing, `husky` throws an error:

```
[node-sdk] $ git commit                                                                                                                                                                                                                             (temp---S)
No .pre-commit-config.yaml file was found
- To temporarily silence this, run `PRE_COMMIT_ALLOW_NO_CONFIG=1 git ...`
- To permanently silence this, install pre-commit with the --allow-missing-config option
- To uninstall pre-commit run `pre-commit uninstall`
```

We should see if there is a way that users can still run `git commit` without needing these files.

dpopp07:
So, the problem I described above only occurs when `pre-commit` is installed for the repo. Since external users won't be installing this, they shouldn't have a problem

PR #898, opened 2019-06-12
fix: use Record instead of Map type for classifier update

evanshortiss:
This fix address a problem with `updateClassifier` related to the `UpdateClassifierParams` interface. The interface was specifying `Map` as the type for `positive_examples`, but it should have been a `Record` based on reading [the code](https://github.com/watson-developer-cloud/node-sdk/blob/2ed11004bb22b3d3e6d2fa8a449f0b290722b057/visual-recognition/v3.ts#L531) since a call to `Object.keys` on a `Map` returns empty array instead of a set of keys.

Using a `Map` causes update calls to to fail with a 500 status since the zip data is not sent in the request body. 

#### Checklist

- [ ] tests are included

I didn't update tests since this is purely a types fix and I didn't spot any specific test to change.

Running tests fails for me with `Jest encountered an unexpected problem`. Here is sample output:

```
    SyntaxError: /workspaces/node-sdk/authorization/v1.ts: Unexpected token (18:11)

      16 | 
      17 | import { BaseService } from 'ibm-cloud-sdk-core';
    > 18 | import url = require('url');
```

Not sure what the cause is. Maybe a missing Jest plugin in package.json?

### DIFF
--- a/visual-recognition/v3.ts
+++ b/visual-recognition/v3.ts
@@ -249,7 +249,7 @@ class VisualRecognitionV3 extends BaseService {
    *
    * @param {Object} params - The parameters to send to the service.
    * @param {string} params.name - The name of the new classifier. Encode special characters in UTF-8.
-   * @param {Map<string, NodeJS.ReadableStream|FileObject|Buffer>} params.positive_examples - A dictionary that contains
+   * @param {Record<string, NodeJS.ReadableStream|FileObject|Buffer>} params.positive_examples - A dictionary that contains
    * the value for each classname. The value is a .zip file of images that depict the visual subject of a class in the
    * new classifier. You can include more than one positive example file in a call.
    *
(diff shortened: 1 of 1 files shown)


visual-recognition/v3.ts
```diff
@@ -482,7 +482,7 @@ class VisualRecognitionV3 extends BaseService {
    *
    * @param {Object} params - The parameters to send to the service.
    * @param {string} params.classifier_id - The ID of the classifier.
-   * @param {Map<string, NodeJS.ReadableStream|FileObject|Buffer>} [params.positive_examples] - A dictionary that
+   * @param {Record<string, NodeJS.ReadableStream|NodeJS.ReadStream|FileObject|Buffer>} [params.positive_examples] - A dictionary that
```
evanshortiss: I also set this to `ReadStream` since `createReadStream` returns one. `ReadStream` seems to inherit from `ReadableStream` and worked in my test code:

```ts
  const updateParams: VisualRecognitionV3.UpdateClassifierParams = {
    classifier_id: IBM_CLASSIFIER_ID,
    positive_examples: {
      'my-example': createReadStream(pathToImagesZip)
    }
  }
```

dpopp07: I'm curious about this change - would you point me to the documentation you read on this?

Does `ReadableStream` _not_ work for what `createReadStream` returns? Maybe we want to just use `ReadStream` instead, I'm not sure we need to have both

evanshortiss:
@dpopp07 just double checked. I was wrong on the ReadStream. `createReadStream` returns a  ReadStream, which is a subclass of `stream.Readable` so it should, and does work. I must have gotten mixed up yesterday when I thought it didn't work.

I've reverted that change since it's the just the `Record` one that's necessary.

dpopp07:
Okay, great. This looks good, I will update our generator and merge this soon.

That said, I do have one more comment. There is one more instance of `Map<>` in the `createClassifier` method. Unless there is a reason that it shouldn't be changed, I think it should be updated as well for consistency.

evanshortiss:
@dpopp07, ah of course! 🤦‍♂ 

All good now

dpopp07:
Perfect, thanks! Will merge today

PR #907, opened 2019-07-11
fix: only set readableObjectMode in recognize-stream if not present

MasterOdin:
##### Checklist

- [ ] tests are included

This fixes a bug in `lib/recognize-stream.ts` where on Node 12 using it would throw the following error:
```
TypeError: Cannot set property readableObjectMode of #<Readable> which has only a getter
    at new RecognizeStream (./node_modules/ibm-watson/lib/recognize-stream.js:126:38)
```

This is because Node 12.3 added a [readableObjectMode](https://nodejs.org/docs/latest-v12.x/api/stream.html#stream_readable_readableobjectmode) getter to the Readable stream prototype. This patch changes it such that it now explicitly sets the `readableObjectMode` option before passing it to the Duplex constructor (which in turn gets passed to the Readable constructor).

It should be noted that on Node 10 and below (which is how it currently works), `this.readableObjectMode` can be `true / undefined` while on Node 12, `this.readableObjectMode` can be `true / false`. I'm not sure if you folks would want to break backwards compatibility (slightly) and make `this.readableObjectMode` be `true / false` to match the new Node 12 signature.

I tested these changes locally in a project I'm working on. No tests were added/changed as no unit tests for the class exists, and I guess it's handled indirectly through the integration tests, but I'm not sure how to set that all up.

### DIFF
--- a/lib/recognize-stream.ts
+++ b/lib/recognize-stream.ts
@@ -133,11 +133,13 @@ class RecognizeStream extends Duplex {
   constructor(options) {
     // this stream only supports objectMode on the output side.
     // It must receive binary data input.
-    super(options);
     if (options.objectMode) {
       options.readableObjectMode = true;
-      this.readableObjectMode = true;
       delete options.objectMode;
+    }
+    super(options);
+    if (options.readableObjectMode && this.readableObjectMode === undefined) {
+      this.readableObjectMode = true;
     }
     this.options = options;
     this.listening = false;


MasterOdin:
Not sure why the integration tests didn't run (which is probably why the code coverage is reported as tanking so much) and not sure if there's anything I can do to get them running on Travis?

dpopp07:
@MasterOdin Thanks for the PR!

About the integration tests, yeah I was just noticing that... That is certainly why the coverage is reporting so low, I do not expect that drop when merged 🙂

The credentials are encrypted and shipped with the repo so I don't know why they wouldn't have run. I'll look into that.

I will review this today 👍

MasterOdin:
Unfortunately, due to the need for encryption of credentials, it's impossible to have that work for PRs from forks (which makes sense as it prevents me from revealing them and stealing them).

It might be helpful to write some words on how to set things up in tests/readme.md, especially the services that require setting them up in some way (e.g Watson Assistant, Discovery)

dpopp07:
That's a good idea, I'll try to look into that. As far as this:

> It should be noted that on Node 10 and below (which is how it currently works), this.readableObjectMode can be true / undefined while on Node 12, this.readableObjectMode can be true / false. I'm not sure if you folks would want to break backwards compatibility (slightly) and make this.readableObjectMode be true / false to match the new Node 12 signature.

What would be an example of a user's code that would be rendered incompatible by this change? I think we've accepted `readableObjectMode: false` and that hasn't caused any issues.

MasterOdin:
Yeah, I think I was a bit ambiguous on what I mean. The change doesn't affect what parameters are accepted by the function, but rather just what `readableObjectMode` field might return on an instantiated `RecognizeStream` object.

For example:
```
let a = new RecognizeStream({});  // or {readableObjectMode: false}
console.log(a.readableObjectMode);
```

On Node 10, this will print `undefined` and on Node 12, this will print `false`, as currently this line:
https://github.com/watson-developer-cloud/node-sdk/blob/1eb80807853ace54fa74347aa08d7370b6ba7ed3/lib/recognize-stream.ts#L139
is the only line in which that property gets set on Node 10.

dpopp07:
@MasterOdin Okay, that makes sense - thanks. I think we hold off for now and I merge this PR as is.

Feel free to open an issue if you think we should change this in v5. Note though that we will still be supporting Node 10 in v5 so we will need a solution that is compatible with both.

Issue #764, opened 2018-08-06
[speech-to-text] Missing types for `recognizeUsingWebSocket` (formerly `createRecognizeStream`)


Waldo000000:
When using the speech to text SDK in TypeScript, it would be helpful if the params object required by `recognizeUsingWebSocket` was typed more specifically than `any` (https://github.com/watson-developer-cloud/node-sdk/blob/master/speech-to-text/v1.ts#L453).

dpopp07:
@Waldo000000 I will implement this change, but it's not my highest priority so it may be some time before I get to it. If you need this done soon, feel free to open a PR and I will take a look at it. Thanks

anweshan:
@dpopp07 if I'm not mistaken the accepted parameters should generally be the same from former `createRecognizeStream` right? I can't find the exact differences but it might help to just have that in comment form the previous accepted parameters to `createRecognizeStream` as accepted in `recognizeUsingWebsocket`

germanattanasio:
What's the status of this issue @dpopp07?

dpopp07:
@germanattanasio This never made it into my backlog but it would be easy - it just involves writing a type definition. I can add this to the work for the next major release.

MasterOdin:
@dpopp07 So looking over `lib/recognize-stream.ts`, should all of these have a concrete definition (defined under top-level `interface Options`), or are they supposed to be optional?
```
  /* Query Params*/
  accessToken: string;
  watsonToken: string;
  model: string;
  languageCustomizationId: string;
  acousticCustomizationId: string;
  baseModelVersion: string;
  xWatsonLearningOptOut: boolean;
  xWatsonMetadata: string;

  /* Opening Message Params */
  contentType: string;
  customizationWeight: number;
  inactivityTimeout: number;
  interimResults: boolean;
  keywords: string[];
  keywordsThreshold: number;
  maxAlternatives: number;
  wordAlternativesThreshold: number;
  wordConfidence: boolean;
  timestamps: boolean;
  profanityFilter: boolean;
  smartFormatting: boolean;
  speakerLabels: boolean;
  grammarName: string;
  redaction: boolean;
  processingMetrics: boolean;
  processingMetricsInterval: number;
  audioMetrics: boolean;
```

And are the last three parameters new and will be added to the API documentation (https://cloud.ibm.com/apidocs/speech-to-text?code=node#recognize-audio-websockets-) when v5 of the SDK is released?

dpopp07:
[...]

Issue #637, opened 2018-02-16
[Conversation]Types are not exported


RoxanaDevIBM:
Hey, 
I am using watson-developer-cloud with a TypeScript project where we use the conversation service.

When writing the code, I noticed that I was not able to use any of the types you guys defined under the namespace ConversationV1 in v1-generated.ts, and was advised to open a ticket.
This means that while using the code, I need to either generate the types myself or use any as a type.

For example:
The following code in TS would cause an error:
```js
const params: ConversationV1.GetWorkspaceParams = {
  workspace_id: workspaceId,
  export: false
};
```

This is caused by the fact that GetWorkspaceParams is something which is defined on the namespace, but not on the class which I import.

I was digging for a couple of solutions:
1. The most easy way would be to add/expose the types to the exported class in v1.d.ts by adding the following tot he class members:
```GetWorkspaceParams: GeneratedConversationV1.GetWorkspaceParams;```
The problem is, that there are a lot of types, and this seems a bit much to add them all, plus this solution does not seem right.
2. Another easy way is to expose v1-generated in index.d.ts, this will allow using it as a namespace.
3. Writing an internal type definition file for exposing the definitions I am using, tho this would cause not needed code duplication.
4. The hard option is to change the structure of v1, tho it lives up to the TypeScript requirements of the definition file generated. It looks like we lose the namespace when we extend the merged class/namespace GeneratedConversationV1, into ConversationV1 class and export that. This throws me back to nr.2.

The reason why I think it would be great to be able to use those types is that the intellisens given by this project is already great, and it would be a shame not to use it. Another reason is that of course when on TypeScript, any should be avoided as much as possible.

Any ideas? :)

**Comment by ammardodin** [comment]
Hey @RoxanaDevIBM thanks for reporting the issue and bring this up to our attention. The team recently switched to `TypeScript`. We will evaluate the options you provided and see what the right approach is here.

For the time being, can you try doing the following ?
```javascript
import{ GetWorkspaceParams } from 'watson-developer-cloud/conversation/v1-generated';
```
I think you can import the types you want to use via something like the above.

anweshan:
So we have investigated the options you suggested - but ultimately we are investigating removing namespaces altogether as namespaces are not recommended in tslint as ES6-style external modules are the standard way to modularize code. For now all users can use the workaround provided and just import whatever types they want from the generated code

germanattanasio:
[...]

Issue #897, opened 2019-06-12
v5 Goals


dpopp07:
## Schedule

We plan to release v5 in September 2019. This issue will be updated with more specifics on timing and the availability of release candidates as we get closer to that time.

8/30/19: [Release Candidate 1](https://github.com/watson-developer-cloud/node-sdk/releases/tag/v5.0.0-rc1)

## Breaking Changes
- [ ] Change how we export types **Seeking community feedback**
- [ ] Find a new request library; `axios` has presented some problems
- [x] Remove `return_response` parameter and always return the detailed response ([896](https://github.com/watson-developer-cloud/node-sdk/issues/896))
- [x] Pass the detailed response as the first argument when using callbacks
- [x] Remove support for Node versions 6 (EOLed in April) and 8 (will EOL in December)
- [x] Use lower-camel-case parameter names to match typical JS style convention
- [x] Decouple authentication mechanism from service constructor
- [x]  Remove location of direct export of the IAM Token Manager to be with the new Authenticators
- [x] Add types to all Promise return values

## Non-breaking major changes
- [x] Add type definitions for WebSocket methods ([764](https://github.com/watson-developer-cloud/node-sdk/issues/764))
- [x] Improve browser compatibility
- [ ] ~Use only TypeDoc for generated documentation~
- [ ] Services can be loaded independently using webpack

---
This issue will be used to track features and changes we want to make in the next major release of this package. We welcome community feedback on these goals - do people want these things? Are there other implementations/packages we should be looking at? Any comments are appreciated!

npacucci:
All right, good, guys. I can't wait for the browser compatibility improvements!

dpopp07:
@npacucci Definitely. Feel free to let us know if you have any suggestions as well!

MasterOdin:
Directed here from #908 about the typing which would be beneficial to address as it causes CI failures for stuff like #910 which attempts to follow current practices.

[...]

PR #908, opened 2019-07-12
refactor: fill out more typing on speech-to-text/v1

MasterOdin:
Fills out typing information on some of the functions that had implicit `any` on their parameters/return. There's still some missing, principally for types that need to be defined in the v1.ts file. For those types, should they be defined as an interface/type under a `SpeechToTextV1` namespace in `v1.ts` or is there some other preferred method? https://github.com/watson-developer-cloud/node-sdk/issues/637#issuecomment-373130130 seems to indicate that you folks were going to move away from namespaces, but not sure the status of that.

The three imports removed were not used at all in the file. There's still two defined conts (`protocols` and `PARAMS_ALLOWED`) and function (`formatChunk`) that still exist.

Related to #764.

##### Checklist

### DIFF
--- a/speech-to-text/v1.ts
+++ b/speech-to-text/v1.ts
@@ -1,9 +1,6 @@
 import async = require('async');
 import extend = require('extend');
-import { getMissingParams } from 'ibm-cloud-sdk-core';
 import isStream = require('isstream');
-import pick = require('object.pick');
-import { parse } from 'url';
 import RecognizeStream = require('../lib/recognize-stream');
 import GeneratedSpeechToTextV1 = require('./v1-generated');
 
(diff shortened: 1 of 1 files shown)


dpopp07:
@MasterOdin This looks good, thanks.

To answer your question, we are likely moving away from namespaces (although the comments in that issue are somewhat dated) and are thinking about refactoring how we expose our custom types in general come v5. If you're a TypeScript user and have any opinions on how we could improve our TS experience, I'd be interested to hear them - best place to do that is [in the v5 goals issue](https://github.com/watson-developer-cloud/node-sdk/issues/897).

For now, if you're wanting to use types in this file, you can add them whichever way you think is best and we can discuss. Were you planning on adding them to this PR? If so, I'll hold off on merging.

MasterOdin:
Alright, thank you for your response. I'll look to use namespaces then for it to keep it consistent with what's there.

Feel free to merge now as it'll probably take me a bit before I've got the rest of the typing done.

PR #909, opened 2019-07-12
ci: only run deploy on watson-developer-cloud/node-sdk repo

MasterOdin:
Pushes to master on forks will fail on Node 8 as it runs the deploy section as it attempts to make a push to this repo. This PR modifies Travis-CI to only run the `deploy` section only if the build is running on watson-developer-cloud/node-sdk.

example: https://travis-ci.com/MasterOdin/node-sdk/jobs/215648162

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@ sudo: false
 node_js:
 - 6
 - 8
+cache:
+  npm: false
 before_install:
 - '[ "${TRAVIS_PULL_REQUEST}" = "false" ] && [ ! -z ${encrypted_ac3aacad7ba8_key} ] &&
   openssl aes-256-cbc -K $encrypted_ac3aacad7ba8_key
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,3 +27,4 @@ deploy:
   script: npx semantic-release
   on:
     node: 8
+    repo: watson-developer-cloud/node-sdk


MasterOdin:
Looked into why Node 6 failed, and it's unfortunately an issue with Travis deals with node builds and semantic-release

Travis does a couple of things for node builds automatically under the hood if you don't specify things explicitly. If you don't specify it to not cache, It will automatically cache the `node_modules` folder for a build and then should restore that on the next build run. Additionally, it will always prepend `./node_modules/bin` to the `$PATH`. Finally, if you don't specify the `install` section in `.travis.yml`, it will automatically run either `npm ci` (if version of npm supports it) or `npm install`.
 
This repo has `semantic-release` as a devDependency that gets installed, which in-turn installs `@semantic-release/npm` which in turn has a dependency on `npm` which then installs the newest version of `npm` (which supports `npm ci`) to `./node_modules/bin/npm`.

Running `npm ci` first deletes everything in `./node_modules`  and then just looks at the `package-lock.json` file to determine what to install.

The sequence then (if a cache is found) is that Travis will run `./node_modules/bin/npm ci` which in turn 
 immediately deletes `./node_modules` which then causes npm to fail when it goes to actually install stuff.

I'm kind of surprised this hasn't actually happened yet and why this was the first couple pushes to start picking up caches. The easiest fix would be to either don't cache node_modules, explicitly delete node_modules in `before_install` section or just explicitly add `install: npm install` to the travis configuration.

I've also filed an issue for Travis: https://travis-ci.community/t/npm-ci-will-fail-if-cached-dependency-includes-npm/4203

dpopp07:
I started looking into the Node 6 failure and discovered the same thing, although I didn't follow my investigation as thoroughly as you did - thanks for posting those details. Thanks for opening that issue with Travis as well. I also don't know why this hasn't happened before, there must have been a change to either npm or Travis.

I think the solution I would prefer is to just not cache the node modules, if that fixes it. That way we still get the benefits of `npm ci`. Caching has caused issues on Travis for this repo in the past anyways.

Do you want to add that to this PR?

MasterOdin:
Looking up the right thing to do to disable it revealed why this is suddenly a thing (https://docs.travis-ci.com/user/caching/#npm-cache):
> Please note that as of July 2019, npm is cached by default on Travis CI

Made the change though to disable the npm cache to go back to the previous working behavior.

dpopp07:
I read the same thing right as you commented - mystery solved! Thanks for making that update

PR #910, opened 2019-07-13
refactor: add typing to authorization/v1

MasterOdin:
Adds typing information to `authorization/v1`. No functionality in the class was changed.

Some notes (possible changes I think for v5):
* Had to disable a tslint rule for the existing class variable `target_url`. Didn't want to change it to pass tslint and break backwards compatibility.
* Should the callback be mandatory for getToken? Currently, it's possible to do `auth.getToken({url: auth.target_url});` which will cause the following error to be shown (if using the `apikey` method at least):
```
(node:30952) UnhandledPromiseRejectionWarning: TypeError: cb is not a function
    at /Users/mpeveler/Work/Github/node-sdk/node_modules/ibm-cloud-sdk-core/auth/jwt-token-manager-v1.js:77:24
    at /Users/mpeveler/Work/Github/node-sdk/node_modules/ibm-cloud-sdk-core/lib/requestwrapper.js:206:13
    at processTicksAndRejections (internal/process/task_queues.js:82:5)
(node:30952) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 2)
(node:30952) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```
which I'd find confusing. Given that it doesn't make sense to not use a callback (or else you don't get the generated token to use for the service), it should probably be mandatory?

##### Checklist

### DIFF
--- a/authorization/v1.ts
+++ b/authorization/v1.ts
@@ -14,13 +14,16 @@
  * limitations under the License.
  */
 
+import { IncomingHttpHeaders } from 'http';
 import { BaseService } from 'ibm-cloud-sdk-core';
 import url = require('url');
 
 class AuthorizationV1 extends BaseService {
   static URL: string = 'https://stream.watsonplatform.net/authorization/api';
-  name;
-  serviceVersion;
+  name: string; // set by prototype to 'authorization'
+  serviceVersion: string; // set by prototype to 'v1'
+  // tslint:disable-next-line:variable-name
+  target_url?: string;
 
   /**
    * Authorization Service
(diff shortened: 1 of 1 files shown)


dpopp07:
Thanks for the PR and the suggestions. I think both of those things would be good changes for v5.

I'll keep those in mind but also, there's a possibility that this module will go away in v5. This would be in favor of just using the token manager directly, since "watson tokens" have pretty much gone away in favor of IAM. We may keep this for convenience though, it's still being discussed. Just a heads up.

MasterOdin:
That's fine as I don't personally use this module. I just wanted to use this as an initial try at adding more elaborate typing before I tackle the larger/more complex classes/functions that lack typing.

PR #916, opened 2019-07-17
docs: add deprecation notice for node versions 6 and 8

dpopp07:
bumping the core version will include the deprecation notice from the base service in the code

also added the notice to the "announcement" section of the readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,8 +49,8 @@ Node.js client library to use the Watson APIs.
 </details>
 
 ## ANNOUNCEMENTS!
-### New Major Version
-The new major version, `v4` has been released. The major new feature in this release is native support for Promises. There are some breaking changes - use the [migration guide](https://github.com/watson-developer-cloud/node-sdk/blob/master/UPGRADE-4.0.md) to navigate any changes that affect your code.
+### Deprecating support for Node versions 6 and 8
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -86,11 +86,11 @@
...
     "dotenv": "^6.2.0",
     "extend": "~3.0.2",
-    "ibm-cloud-sdk-core": "^0.3.1",
+    "ibm-cloud-sdk-core": "^0.3.3",
     "isstream": "~0.1.2",
     "object.pick": "~1.3.0",
-    "websocket": "^1.0.28",
-    "snyk": "^1.192.4"
+    "snyk": "^1.192.4",
+    "websocket": "^1.0.28"
```
dpopp07: This just re-ordered to be alphabetical. `npm` does that automatically

README.md
```diff
@@ -49,8 +49,8 @@ Node.js client library to use the Watson APIs.
 </details>
 
 ## ANNOUNCEMENTS!
-### New Major Version
-The new major version, `v4` has been released. The major new feature in this release is native support for Promises. There are some breaking changes - use the [migration guide](https://github.com/watson-developer-cloud/node-sdk/blob/master/UPGRADE-4.0.md) to navigate any changes that affect your code.
+### Deprecating support for Node versions 6 and 8
```
germanattanasio: I think that we can just say that we support Node > 10. No need to say that we deprecated the support for an older version unless we know it won't work

dpopp07: Sounds good to me 👍

PR #918, opened 2019-07-18
Drop support for node 6 and 8

dpopp07:
Officially drop support for Node versions 6 and 8 in the `next` branch of the SDK.


UPGRADE-5.0.md
```diff
@@ -0,0 +1,7 @@
+# Upgrading to ibm-watson@4.0
```
germanattanasio: Move the `UPGRADE-4.0.md` to a wiki page and add a line in this file saying that if your version is prior to 4.0 look at that wiki

dpopp07: Good idea 👍

PR #924, opened 2019-08-02
Node SDK Dockerfile

kingsodigie:
Added docker file to sdk which when built spins up an image with the node-sdk installed.


docker/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+#Install Git
+RUN apt-get update && \
+    apt list -a git && \
+    apt-get install -y --no-install-recommends git=1:2.*
+
+#Clone SDK repo to app directory
+RUN git clone https://github.com/watson-developer-cloud/node-sdk.git
+
+#Install the SDK
+RUN npm install ibm-watson
```
dpopp07: If we are building the SDK from source, why install it from NPM as well? Or vice versa? It seems like we would just need one or the other.

I'm no Docker expert so I guess I'm just curious to hear your reasoning!

docker/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+#Set working directory to node-sdk
+WORKDIR /node-sdk
+
+#Install dependencies and build to run tests.
+RUN npm install && \
+    npm run build
+
+#Run tests to verify installation
+CMD ["npm" , "test"]
```
dpopp07: This is just a suggestion - `npm test` runs the full suite of integration tests, which have a variable (and unfortunately high) failure rate due to service timeouts, etc. They also take ~25 minutes to run on average. It may be better to run only the unit tests:

`CMD ["npm" , "run" , "test-unit"]`

docker/Dockerfile
kingsodigie: I build the SDK from the source in order to be able to verify the installation via the test cases. I was unsure if it was possible to just build the test cases and not install the SDK from the source and rather only install the SDK with npm as the original intention was to install the SDK as a "user" rather than a "developer".

dpopp07: Hmm I see what you were going for, but I'm not sure it's necessary. Running `npm install ibm-watson` without any failures implies a successful installation - that's how we expect users to install the package in the wild. Downloading the source and running the tests just makes sure that the cloning of the repo was successful - it's independent of the package install from `npm`.

If you really wanted to double-check that the SDK is ready and usable after `npm install`, you could write a script that checked for the presence of the files in the `node_modules/` folder. Or, you could include a JS file that uses that imports the SDK but doesn't run any operations, then make sure that runs without error.

docker/Dockerfile
```diff
@@ -0,0 +1,8 @@
+#Import Node base image(change version as needed)
```
dpopp07:
```suggestion
# Import Node base image (change version as needed)
```

Just being annoying about spacing 😛

docker/Dockerfile
```diff
@@ -0,0 +1,8 @@
...
+FROM node:12-slim
+
+#Install the node SDK
+RUN npm install ibm-watson
+
+
+
```
dpopp07: Will you remove the extra white space at the end of the file?

docker/Dockerfile
```diff
@@ -0,0 +1,5 @@
+# Import Node base image (change version as needed)
+FROM node:12-slim
+
+#Install the node SDK
+RUN npm install ibm-watson
```
dpopp07: Still need a newline at end of file

```suggestion
RUN npm install ibm-watson

```

PR #928, opened 2019-08-07
fix(core): share request config across all requests

dpopp07:
* fix added by bumping core to 0.3.5, which includes the important changes
* this will allow users to configure all requests, which fixes a bug that prevented use of the SDK behind a corporate proxy
* documentation added for this use case in the readme

Also, removed some read docs from the readme - couldn't fine the `setServiceUrl` method anywhere in this SDK or the core.


dpopp07:
@germanattanasio gave his blessing, merging

PR #930, opened 2019-08-08
fix: send correct user-agent header for websocket methods

dpopp07:
I recently noticed that the analytics headers for the WebSocket methods do not follow the new format. Making this quick change to correct them

### DIFF
--- a/text-to-speech/v1.ts
+++ b/text-to-speech/v1.ts
@@ -1,8 +1,7 @@
 import extend = require('extend');
+import { getSdkHeaders } from '../lib/common';
 import SynthesizeStream = require('../lib/synthesize-stream');
 import GeneratedTextToSpeechV1 = require('./v1-generated');
-// tslint:disable-next-line:no-var-requires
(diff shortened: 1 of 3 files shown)


text-to-speech/v1.ts
```diff
@@ -87,11 +86,13 @@ class TextToSpeechV1 extends GeneratedTextToSpeechV1 {
       params.token_manager = this.tokenManager;
     }
 
+    // include analytics headers
+    const sdkHeaders = getSdkHeaders('text_to_speech', 'v1', 'synthesizeUsingWebSocket');
```
germanattanasio: Will this break the browser users?

dpopp07: Hmm I'm not sure. We were still sending the `user-agent` header before but not the `x-ibm-cloud-analytics` header so it depends on if that will cause CORS issues.

That said, these two operations are the ones we have a completely separate browser package for so I don't know if it's a big deal 🤷‍♂

PR #946, opened 2019-08-29
feat: add support for new authenticators in all sdks

dpopp07:
- All service regenerated
- Core version updated to use `release-candidate-v1` branch
- Authenticators required (and exported from the watson sdk)
- All parameters changed to lower camel case
- Only detailed response is returned

Many of the breaking changes come just from the core version changing.

All breaking changes described in the Migration Guide.

I will release the RC once this is merged.


package.json
```diff
@@ -87,7 +87,7 @@
     "camelcase": "^5.3.1",
     "dotenv": "^6.2.0",
     "extend": "~3.0.2",
-    "ibm-cloud-sdk-core": "^0.3.5",
+    "ibm-cloud-sdk-core": "github:IBM/node-sdk-core#release-candidate-v1",
```
mediumTaj: pretty sweet... for the RC for the SDK will you do an RC for the core to replace this?

PR #953, opened 2019-09-13
fix: pass user-defined http(s) agent to websocket methods

dpopp07:
We accept `axios` options in the constructor to give the user extended control over their requests but we didn't translate this control over to the WebSocket requests. This PR doesn't give _complete_ control over the WebSocket configuration but it allows the use of custom "agents" which will allow users to use WebSockets behind corporate proxies (an established use case).


lib/recognize-stream.ts
```diff
@@ -241,12 +242,18 @@ class RecognizeStream extends Duplex {
...
+    const { agent } = options;
+    const requestOptions = agent ? { agent } : null;
+
     const socket = (this.socket = new w3cWebSocket(
       url,
       null,
       null,
       options.headers,
-      null,
+      requestOptions,
```
germanattanasio: I guess the change here is based on https://github.com/theturtle32/WebSocket-Node/blob/master/docs/WebSocketClient.md#connect-using-a-proxy-server

dpopp07: Yep, that's correct!

PR #954, opened 2019-09-13
refactor: add typing to all non-generated functions of speech-to-text

MasterOdin:
This adds typing for the three functions of `./speech-to-text/v1.ts` that lacked them, as well as removes any implicit definitions of any in the file.

I decided it would be more pragmatic to make `RecognizeStream.Options` a superset of `Duplex.Options` and `SpeechToTextV1.RecognizeWebSocketParams` and not duplicate the definitions across the two files, especially given the stream is only used for that particular method in the SDK itself.

I also added the optional bit to all optional properties of `RecognizeStream.Options` as part of this.

I used `extend` a couple of times (as it was already in the project and in this file) so that I wouldn't have to make (for example) `WhenCorporaAnalyzedOptions.errorFilter` "optional" to allow it to be defined on the next line. I didn't do that with `SpeechToTextError` as it's not possible to define a second class in the file per the tslint and so had to make due with a weaker interface design with optional `code` even though it should never actually be undefined in usage.

##### Checklist

### DIFF
--- a/lib/recognize-stream.ts
+++ b/lib/recognize-stream.ts
@@ -14,62 +14,21 @@
  * limitations under the License
  */
 
-import { OutgoingHttpHeaders } from 'http';
 import { Authenticator, contentType, qs } from 'ibm-cloud-sdk-core';
 import omit = require('object.omit');
 import pick = require('object.pick');
 import { Duplex, DuplexOptions } from 'stream';
 import { w3cwebsocket as w3cWebSocket } from 'websocket';
+import SpeechToTextV1 = require('../speech-to-text/v1-generated');
 import { processUserParameters } from './websocket-utils';
 
-// these options represent the superset of the base params,
-// query params, and opening message params, with the keys
-// in lowerCamelCase format so we can expose a consistent style
-// to the user. this object should be updated any time either
-// openingMessageParamsAllowed or queryParamsAllowed is changed
-interface Options extends DuplexOptions {
-  /* Base Properties */
-  authenticator: Authenticator;
-  url?: string;
-  headers?: OutgoingHttpHeaders;
-  readableObjectMode?: boolean;
-  objectMode?: boolean;
-  disableSslVerification?: boolean;
-
-  /* Query Params*/
-  accessToken: string;
-  watsonToken: string;
-  model: string;
-  languageCustomizationId: string;
-  acousticCustomizationId: string;
-  baseModelVersion: string;
-  xWatsonLearningOptOut: boolean;
-  xWatsonMetadata: string;
-
-  /* Opening Message Params */
-  contentType: string;
-  customizationWeight: number;
-  inactivityTimeout: number;
-  interimResults: boolean;
-  keywords: string[];
-  keywordsThreshold: number;
-  maxAlternatives: number;
-  wordAlternativesThreshold: number;
-  wordConfidence: boolean;
-  timestamps: boolean;
-  profanityFilter: boolean;
-  smartFormatting: boolean;
-  speakerLabels: boolean;
-  grammarName: string;
-  redaction: boolean;
-  processingMetrics: boolean;
-  processingMetricsInterval: number;
-  audioMetrics: boolean;
+interface WritableState {
+  highWaterMark: number;
 }
 
 interface RecognizeStream extends Duplex {
-  _writableState;
-  readableObjectMode;
+  _writableState: WritableState;
+  readableObjectMode: boolean;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


MasterOdin:
Codacy quality review is on adding the `namespace RecognizeStream` and `namespace SpeechToTextV1`, not sure if there was a preferred new method yet on how to do this typing as I believe all other classes still use this method.

dpopp07:
@MasterOdin I actually couldn't care less about what Codacy says, a lot of times it even contradicts our own style/linting guides. That said, I was going to comment on the introduction of namespaces here since I believe they are [generally recommended against](https://palantir.github.io/tslint/rules/no-namespace/) and I've been thinking about trying to refactor how we organize the other files to remove them.

Now I see why you would want them for the STT class so that it's consistent for the other services but for the RecognizeStream class, would it make a big difference to do something like this?

```ts
export interface RecognizeStreamOptions extends DuplexOptions, SpeechToTextV1.Options {
  // these options represent the superset of the base params,
  // query params, and opening message params, with the keys
  // in lowerCamelCase format so we can expose a consistent style
  // to the user.
  authenticator: Authenticator;
  url?: string;
}
```

lib/recognize-stream.ts
```diff
@@ -92,12 +51,12 @@ class RecognizeStream extends Duplex {
...
   }
 
-  private options: Options;
+  private options: RecognizeStream.Options;
   private authenticator: Authenticator;
   private listening: boolean;
   private initialized: boolean;
   private finished: boolean;
-  private socket;
+  private socket: w3cWebSocket;
```
dpopp07: Huh, I thought I added this before and got compiler issues. I guess I was doing something different. Glad to see this 👍

lib/recognize-stream.ts
```diff
@@ -14,62 +14,21 @@
  * limitations under the License
  */
 
-import { OutgoingHttpHeaders } from 'http';
 import { Authenticator, contentType, qs } from 'ibm-cloud-sdk-core';
 import omit = require('object.omit');
 import pick = require('object.pick');
 import { Duplex, DuplexOptions } from 'stream';
 import { w3cwebsocket as w3cWebSocket } from 'websocket';
+import SpeechToTextV1 = require('../speech-to-text/v1-generated');
```
dpopp07: Is there a reason you're pulling this import from the generated module and not the adapter module?

MasterOdin:
Unfortunately, `RecognizeStream`, like all the other classes does at the bottom:
```
export = RecognizeStream;
```
which essentially binds this to using the `namespace` method. You could refactor it so that both the `RecognizeStream` and `RecognizeStreamOptions` are top level exports, but I think runs into the stuff I talk about in https://github.com/watson-developer-cloud/node-sdk/issues/897#issuecomment-511161443. Might be better to handle this class at the same time as all the others so that they all share the same structure?

speech-to-text/v1.ts
```diff
@@ -225,4 +231,66 @@ class SpeechToTextV1 extends GeneratedSpeechToTextV1 {
   }
 }
 
+namespace SpeechToTextV1 {
+  export type Callback<T> = (err: Error | SpeechToTextError | null, res?: T) => void;
+
+  export interface SpeechToTextError extends Error {
+    message: string;
+    code?: string;
```
dpopp07: > had to make due with a weaker interface design with optional code even though it should never actually be undefined in usage.

I think that's fine. It can always be revisited in the future but I think that's a very small issue

dpopp07:
> Might be better to handle this class at the same time as all the others so that they all share the same structure?

That's a good point, we can do that.

> I do agree in getting rid of the namespaces at some point though to be following TS best practices, and would be happy to help on that front / mock up what a preferred approach might look like.

That's something I've been wanting to get to recently but just have not had the time. If you want to take a stab at the design for that, that would be great and I'd be happy to review it with you. I think the big limitation there is that we are pretty ingrained with the import naming structure for the SDK (`ibm-watson/speech-to-text/v1`, etc.) but we should think about the possibilities.

Another thing to think about is that if it would help to define our own `.d.ts` files (I don't have much experience in that area, specifically) but you think that would be too tedious - we can auto-generate them like we do the rest of code. Just something to be aware of.

lib/recognize-stream.ts
MasterOdin: The autocomplete giveth, the autocomplete taketh away.

Fixed, and used the proper class for the inheritance on `RecognizeStream.Options`.

MasterOdin:
You'd only want to do the `*.d.ts` route only because the library was written in JS and that was the only way to get typings for TypeScript to use (it's the basis of [@types](https://github.com/DefinitelyTyped/DefinitelyTyped)). I can certainly play around with some options  after I've knocked these next two things I was planning on doing out:
* updated husky and commitlint to get rid of the last `7` npm audits (ran into a snag as tsc suddenly started complaining about a bunch of errors, but I think that was just a weird `node_modules` installation)
* added typings for text-to-speech/v1

PR #957, opened 2019-09-16
feat: separate strings out of primary SynthesizeStream pipe

### LINKED ISSUE #956: synthesizeUsingWebSocket producing bad files
Text to speech using synthesizeUsingWebSocket apparently works but the output file can't be opened by any player. The files produced using synthesize (without web socket) work fine but I can't always use the http API because texts are too long.

This is the code:

```javascript
"use strict";
require('dotenv').config();
const fs = require('fs');
const path = require('path');
const TextToSpeechV1 = require('ibm-watson/text-to-speech/v1');
const textToSpeech = new TextToSpeechV1({
	iam_apikey: process.env.APIKEY,
	url: process.env.URL
});
const files = process.argv.slice(2);

function toSpeech(text, fileName) {
	let audio = textToSpeech.synthesizeUsingWebSocket({text, voice: 'es-ES_EnriqueV3Voice', accept: 'audio/ogg;codecs=opus'});
	audio.pipe(fs.createWriteStream(`${fileName}.ogg`));
}

for (let file of files)
	toSpeech(fs.readFileSync(file, 'utf8'), path.basename(file, path.extname(file)));
```

**SDK Version**
4.4.0

**Additional information:**
- OS: Linux Mint 19.2
- Which version of `Node` are you using? 10.16.3

MasterOdin:
Fixes #956 

As discussed in the issue, the SythesizeStream is currently outputting all information, be it strings or binary data to the pipe (after converting to a Buffer). However, following the [example](https://github.com/watson-developer-cloud/node-sdk/blob/master/examples/text_to_speech_websocket.js), this will lead to a file that is invalid as it will append both the text and binary to the same file.

This PR separates out the information, and outputs the information onto their own relevant channel as documented for the different types of information strings on the documentation:
* https://cloud.ibm.com/docs/services/text-to-speech?topic=text-to-speech-usingWebSocket
* https://cloud.ibm.com/docs/services/text-to-speech?topic=text-to-speech-timing
(the bottom of the second demonstrates also that the strings may come intermixed with the binary audio information)

The various things documented as being possible emitted:
* [audio format](https://cloud.ibm.com/docs/services/text-to-speech?topic=text-to-speech-usingWebSocket#WSreceive)
* [error/warnings](https://cloud.ibm.com/docs/services/text-to-speech?topic=text-to-speech-usingWebSocket#returnErrors)
* [marks](https://cloud.ibm.com/docs/services/text-to-speech?topic=text-to-speech-timing#mark)
* [words](https://cloud.ibm.com/docs/services/text-to-speech?topic=text-to-speech-timing#timingRequest)

While this is a breaking change, as it's breaking things to be more expected and inline with how the API is thought to work, putting it as a minor release instead of a major release, which also does not require any documentation change.

No idea how to include a test for this change as well since it looks like the existing test does not directly test the data piped out of the socket.

##### Checklist

Note: Labeled this a `feat` instead of `fix` so that it gets a minor release instead of a bugfix by semantic-release, which I think is probably more appropriate, but can rename the commit if desired.

### DIFF
--- a/lib/synthesize-stream.ts
+++ b/lib/synthesize-stream.ts
@@ -52,6 +52,7 @@ interface SynthesizeStream extends Readable {
  */
 class SynthesizeStream extends Readable {
 
+  static WEBSOCKET_ERROR: string = 'WebSocket error';
   static WEBSOCKET_CONNECTION_ERROR: string = 'WebSocket connection error';
 
   private options;
@@ -142,7 +143,11 @@ class SynthesizeStream extends Readable {
             self.emit('words', message, json);
           }
           else if (json['error']) {
-            self.emit('error', message, json);
+            // this should have same structure as onerror emit
+            const err = new Error(json['error']);
+            err.name = SynthesizeStream.WEBSOCKET_ERROR;
+            err['event'] = message;
+            self.emit('error', err);
           }
           else if (json['warnings']) {
             self.emit('warnings', message, json);
--- a/lib/synthesize-stream.ts
+++ b/lib/synthesize-stream.ts
@@ -68,7 +68,7 @@ class SynthesizeStream extends Readable {
    * Note that the WebSocket connection is not established until the first chunk of data is recieved. This allows for IAM token request management by the SDK.
    *
    * @param {Object} options
-   * @param {String} options.text - The text that us to be synthesized. Provide plain text or text that is annotated with SSML. SSML input can include the SSML <mark> element. Pass a maximum of 5 KB of text. 
+   * @param {String} options.text - The text that us to be synthesized. Provide plain text or text that is annotated with SSML. SSML input can include the SSML <mark> element. Pass a maximum of 5 KB of text.
    * @param {String} options.accept - The requested audio format (MIME type) of the audio.
    * @param {String[]} [options.timings] - An array that specifies whether the service is to return word timing information for all strings of the input text
    * @param {String} [options.voice='en-US_MichaelVoice'] - The voice that is to be used for the synthesis.
@@ -98,7 +98,7 @@ class SynthesizeStream extends Readable {
 
     const url =
       (options.url || 'wss://stream.watsonplatform.net/text-to-speech/api')
-        .replace(/^http/, 'ws') + 
+        .replace(/^http/, 'ws') +
         '/v1/synthesize?' +
         queryString;
 
@@ -126,9 +126,34 @@ class SynthesizeStream extends Readable {
 
     socket.onmessage = message => {
       const chunk = message.data;
-      // some messages are strings - emit those unencoded, but push them to
-      // the stream as binary
-      const data = typeof chunk === 'string' ? chunk : Buffer.from(chunk);
+      // some info messages are sent as strings, telling the content_type and
+      // timings. Emit them as separate events, but do not send them along the
+      // pipe.
+      if (typeof chunk === 'string') {
+        try {
+          const json = JSON.parse(chunk);
+          if (json['binary_streams']) {
+            self.emit('binary_streams', message, chunk);
+          }
+          else if (json['marks']) {
+            self.emit('marks', message, chunk);
+          }
+          else if (json['words']) {
+            self.emit('words', message, chunk);
+          }
+          else if (json['error']) {
+            self.emit('error', message, chunk);
+          }
+          else if (json['warnings']) {
+            self.emit('warnings', message, chunk);
+          }
+        }
+        finally {
+          self.emit('message', message, chunk);
+        }
+        return;
+      }
+
       /**
        * Emit any messages received over the wire, mainly used for debugging.
        *
@@ -136,7 +161,7 @@ class SynthesizeStream extends Readable {
        * @param {Object} message - frame object received from service
        * @param {Object} data - a data attribute of the frame that's either a string or a Buffer/TypedArray
        */
-      self.emit('message', message, data);
+      self.emit('message', message, chunk);
       self.push(Buffer.from(chunk));
     };
 
--- a/lib/synthesize-stream.ts
+++ b/lib/synthesize-stream.ts
@@ -133,19 +133,19 @@ class SynthesizeStream extends Readable {
         try {
           const json = JSON.parse(chunk);
           if (json['binary_streams']) {
-            self.emit('binary_streams', message, chunk);
+            self.emit('binary_streams', message, json);
           }
           else if (json['marks']) {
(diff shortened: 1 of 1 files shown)


lib/synthesize-stream.ts
```diff
@@ -126,17 +126,42 @@ class SynthesizeStream extends Readable {
...
       const chunk = message.data;
-      // some messages are strings - emit those unencoded, but push them to
-      // the stream as binary
-      const data = typeof chunk === 'string' ? chunk : Buffer.from(chunk);
+      // some info messages are sent as strings, telling the content_type and
+      // timings. Emit them as separate events, but do not send them along the
+      // pipe.
+      if (typeof chunk === 'string') {
+        try {
+          const json = JSON.parse(chunk);
```
MasterOdin: technically, I think if it's a string, it should always be a JSON object so the try/finally might not be needed, which would allow for some simplification of this logic (and the two lines following it). Let me know how defensive you think this should be.

MasterOdin: I also figured it was more appropriate to try and separate things out as much as possible instead of just outputting all strings to a user and letting them figure things out, but can condense if desired.

lib/synthesize-stream.ts
```diff
@@ -126,17 +126,42 @@ class SynthesizeStream extends Readable {
...
+          }
+          else if (json['error']) {
+            self.emit('error', message, chunk);
+          }
+          else if (json['warnings']) {
+            self.emit('warnings', message, chunk);
+          }
+        }
+        finally {
+          self.emit('message', message, chunk);
```
MasterOdin: I wanted to keep things consistent with the prior method of how things were outputted, but for the individual channels, may make more sense to emit like `self.emit('<channel>', json, message);`

lib/synthesize-stream.ts
dpopp07: I think it's good to be this defensive here. It's JavaScript after all, plus we don't have control over what comes back from the service. Nice to know this won't choke on anything weird

lib/synthesize-stream.ts
dpopp07: I hear what you're saying but I think we should keep this consistent for now. Returning the whole response in the first arg is flexible since we will probably always want to return that. I don't think it would be a bad idea to pass `json` instead of `chunk` here, though, to give the user an already-parsed object.

lib/synthesize-stream.ts
```diff
@@ -126,17 +126,42 @@ class SynthesizeStream extends Readable {
...
+
       /**
        * Emit any messages received over the wire, mainly used for debugging.
        *
        * @event SynthesizeStream#message
        * @param {Object} message - frame object received from service
        * @param {Object} data - a data attribute of the frame that's either a string or a Buffer/TypedArray
        */
-      self.emit('message', message, data);
+      self.emit('message', message, chunk);
```
dpopp07: Technically, this will be a different result because before, `data` would have been `Buffer.from(chunk)`. Is there a reason we wouldn't want to return this converted to a Buffer? As in, is this intended to be an improvement in behavior? If not, I would want to keep the behavior the same

lib/synthesize-stream.ts
```diff
@@ -126,17 +126,42 @@ class SynthesizeStream extends Readable {
...
+            self.emit('binary_streams', message, chunk);
+          }
+          else if (json['marks']) {
+            self.emit('marks', message, chunk);
+          }
+          else if (json['words']) {
+            self.emit('words', message, chunk);
+          }
+          else if (json['error']) {
+            self.emit('error', message, chunk);
```
dpopp07: This is the only one that makes me uneasy because the other existing error event has one arg, which is an Error object that's been formatted a little:
```ts
    socket.onerror = event => {
      const err = new Error('WebSocket connection error');
      err.name = SynthesizeStream.WEBSOCKET_CONNECTION_ERROR;
      err['event'] = event;
      self.emit('error', err);
      self.push(null);
    };
```

it is very likely that users already have handlers set up for the Error event expecting this type of object. I think it would be prudent here to either a) rename this event to something unique or b) try to put together a similarly formatted Error object to send back as the first arg. There are pros and cons to both - what do you think?

MasterOdin: I completely missed this, thanks for pointing it out. I think just joining them into one common event to listen for is probably best? The only thing missing for these types of errors would be a `name` to attribute to them, but it just be use add `WEBSOCKET_ERROR = 'Websocket error'` to the class definition and then rewrite this bit as:
```
json.name = SynthesizeStream.WEBSOCKET_ERROR;
json.event = message;
self.emit('error', err);
```

MasterOdin: Alright, so played around a bit more with this and it looks like for things like messing up the codec or forgetting `text` argument, the JSON object is sent over the `onmessage` channel to the user and then with an additional message on the `onclose` that says "see previous message for the error details". I'd probably lean towards making them emit on the same channel (and unify warnings to a similar emit format, though on a different channel), as the end result is all the same regardless of the exact reason.

lib/synthesize-stream.ts
MasterOdin: Agreed, I've updated the messages that could emit `json` to send that instead, and that all will output a string on `message` to retain BC.

lib/synthesize-stream.ts
MasterOdin: 👍

lib/synthesize-stream.ts
MasterOdin: You are right, I mistakenly thought it would always send either a string or Buffer, but it sends a `ArrayBuffer` in actuality. I've fixed this so that these two lines will get a `Buffer.from(chunk)` like it was before.

lib/synthesize-stream.ts
dpopp07: That sounds good to me, thanks for looking into it

dpopp07: Everything else looks good. Once you update the error handler, I'll approve and merge 👍

MasterOdin: Alright, updated the error handler to look like the `onerror` signature. I did not touch the `warnings` one as I'm not sure there's a good way to mimic the warnings as JS does not really have a good concept of warnings like [python](https://docs.python.org/3/library/warnings.html#module-warnings) unfortunately.

dpopp07: Yeah I think that's fine and there's no existing API there so I'm not worried about it

PR #966, opened 2019-09-20
chore: remove unused object.pick and objec.emit dependencies

MasterOdin:
Removes the `object.pick` dependency from package.json and cleans up `object.emit` and `extend` from two files that were not using it.

`lib/{recognize,synthesize}-stream.ts` stopped using `object.pick` in aee5a816e50e1f6b36fac2124a44b1798879f66d
`lib/websocket-util.ts` stopped using `extend` in 96901cfb86e178d770b72d2cf0b1dde5750cfc4d

`object.emit` is not even in package.json, but since it wasn't referenced, `tsc` was just auto-removing it when building the file so that was not causing any sort of build errors (I think).

##### Checklist

### DIFF
--- a/lib/recognize-stream.ts
+++ b/lib/recognize-stream.ts
@@ -16,8 +16,6 @@
 
 import { RequestOptions } from 'http';
 import { Authenticator, contentType, qs } from 'ibm-cloud-sdk-core';
-import omit = require('object.omit');
-import pick = require('object.pick');
 import { Duplex, DuplexOptions } from 'stream';
 import { w3cwebsocket as w3cWebSocket } from 'websocket';
 import SpeechToTextV1 = require('../speech-to-text/v1');
--- a/lib/synthesize-stream.ts
+++ b/lib/synthesize-stream.ts
@@ -16,7 +16,6 @@
 
 import { Agent, OutgoingHttpHeaders, RequestOptions } from 'http';
 import { Authenticator, qs } from 'ibm-cloud-sdk-core';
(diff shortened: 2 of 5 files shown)


MasterOdin:
I believe `tslint` deprecated their rule to do this, though `tslint` overall is deprecated and its suggested to switch to https://github.com/typescript-eslint/typescript-eslint instead which I think has the similar rule which is not deprecated.

Could also just make it a compile time error:
```
    "noUnusedLocals": true,                /* Report errors on unused locals. */
    "noUnusedParameters": true,            /* Report errors on unused parameters. */
```

For reference, turning them on, I got:
```
authorization/v1.ts:17:31 - error TS6133: 'OutgoingHttpHeaders' is declared but its value is never read.

17 import { IncomingHttpHeaders, OutgoingHttpHeaders } from 'http';
                                 ~~~~~~~~~~~~~~~~~~~

lib/recognize-stream.ts:384:17 - error TS6133: 'encoding' is declared but its value is never read.

384   _write(chunk, encoding, callback): void {
                    ~~~~~~~~

lib/synthesize-stream.ts:70:11 - error TS6133: 'socket' is declared but its value is never read.

70   private socket;
```

dpopp07:
Yeah I've read about that switch but haven't gotten around to migrating yet. What I'm actually planning to do is integrate `prettier` into `tslint` to get some of the rules we currently get in the JS code into the TS code. But there are other formatting issues we need to solve before we do that. Overall, it's not too high of a priority

PR #968, opened 2019-09-23
refactor: add typings for tts/v1

MasterOdin:
Adds typings for `text-to-speech/v1.ts`, using a similar approach as the `speech-to-text/v1.ts` typings in that the Stream Options are split across the two files as appropriate.

##### Checklist

### DIFF
--- a/lib/synthesize-stream.ts
+++ b/lib/synthesize-stream.ts
@@ -14,38 +14,13 @@
  * limitations under the License
  */
 
-import { Agent, OutgoingHttpHeaders, RequestOptions } from 'http';
+import { Agent, RequestOptions } from 'http';
 import { Authenticator, qs } from 'ibm-cloud-sdk-core';
 import { Readable, ReadableOptions } from 'stream';
 import { w3cwebsocket as w3cWebSocket } from 'websocket';
+import TextToSpeechV1 = require('../text-to-speech/v1');
 import { processUserParameters } from './websocket-utils';
 
-// these options represent the superset of the base params,
-// query params, and opening message params, with the keys
-// in lowerCamelCase format so we can expose a consistent style
-// to the user. this object should be updated any time either
-// payloadParamsAllowed or queryParamsAllowed is changed
-interface Options extends ReadableOptions {
-  /* base options */
-  authenticator: Authenticator;
-  url?: string;
-  headers?: OutgoingHttpHeaders;
-  disableSslVerification?: boolean;
-  agent?: Agent;
-
-  /* payload options */
-  text: string;
-  accept: string;
-  timings?: string[];
-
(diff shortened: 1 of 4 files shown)


text-to-speech/v1.ts
```diff
@@ -70,31 +71,51 @@ class TextToSpeechV1 extends GeneratedTextToSpeechV1 {
...
-    // pass the Authenticator to the SynthesizeStream object
-    params.authenticator = this.getAuthenticator();
-
-    // if the user configured a custom https client, use it in the websocket method
-    // let httpsAgent take precedence, default to null
-    params.agent = this.baseOptions.httpsAgent || this.baseOptions.httpAgent || null;
+  synthesizeUsingWebSocket(params: TextToSpeechV1.SynthesizeWebSocketParams): SynthesizeStream {
+    const streamParams: SynthesizeStream.Options = extend(
+      params,
+      {},
```
MasterOdin: This should maintain the prior desired behavior of `params || {}` without typescript complaining about how {} is not assignable to WebSocketParams or Options due to messing required keys:
```
$ node
> const extend = require('extend');
> extend(null, {});
{}
> extend({}, null);
{}
> extend({}, 'aa')
{ '0': 'a', '1': 'a' }
> extend('aa', {})
{}
> extend('', {})
{}
> extend({}, '')
{}
```

lib/synthesize-stream.ts
```diff
@@ -14,38 +14,13 @@
  * limitations under the License
  */
 
-import { Agent, OutgoingHttpHeaders, RequestOptions } from 'http';
+import { Agent, RequestOptions } from 'http';
 import { Authenticator, qs } from 'ibm-cloud-sdk-core';
 import { Readable, ReadableOptions } from 'stream';
 import { w3cwebsocket as w3cWebSocket } from 'websocket';
+import TextToSpeechV1 = require('../text-to-speech/v1');
```
dpopp07: One minor thing here, and we should probably do this in `recognize-stream` as well, I would prefer this to use object deconstruction so it's really obvious what we're importing.

```suggestion
import { SynthesizeWebSocketParams } = require('../text-to-speech/v1');
```

Since we're not importing the whole service for anything, this makes it more clear to me why the import is needed up front.

text-to-speech/v1.ts
dpopp07: I like this - would you mind making a similar change in the STT code for consistency?

MasterOdin: Sure, done!

lib/synthesize-stream.ts
MasterOdin: Works for me, updated the code here and in RecognizeStream.

PR #972, opened 2019-09-27
refactor(typing): add @types/async and fix typing for its usage

MasterOdin:
This adds `@types/async` and fixes an incompatible type that was being passed to it (it requires a concrete usage of `interval` and `times`, which is provided via the use of `extend` where the CheckParams interface is used.

##### Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1630,6 +1630,11 @@
         "@types/node": "*"
       }
     },
+    "@types/async": {
+      "version": "2.4.2",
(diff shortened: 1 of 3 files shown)


germanattanasio:
@MasterOdin Thanks for the PR, I tried to look you up and found that your website is down `http://www.mpeveler.com`.

MasterOdin:
@germanattanasio Yeah, I keep meaning to find a day to re-setup my entire server after I borked it which had my site. In the meantime, I've put up a static site based out of github to make https://mpeveler.com accessible again.

Issue #984, opened 2019-10-11
Unable to upload zip files in Buffer format


henhen87:
Hello,

I am using the Node-SDK for ibm-watson and am currently trying to Create Classifiers and also compare files with images being uploaded. In the docs here 

[Node VR](https://cloud.ibm.com/apidocs/visual-recognition/visual-recognition-v3?code=node#create-a-classifier)

It states that acceptable formats are NodeJS readable stream or Buffer. When I upload a zip file of images from the front end, the logs in the server (express js server) show that we do indeed have a buffer, which looks something like this:

`<Buffer ff d8 ff e0 00 10 4a 46 49 46 00 01 01 00 00 01 00 01 00 00 ff db 00 43 00 05 03 04 04 04 03 05 04 04 04 05 05 05 06 07 0c 08 07 07 07 07 0f 0b 0b 09 ... >`

When I convert this to a string, I can see the image extensions .jpg/png after encoded strings like this

> ...\u0000\u0000�L\u0004\u0000Panasonic-ES8103S-Nanotech-Electric-Shaver-img3.jpgPK\u0001\u0002\u0014\u0000\u0014\u0000\u0000\u0000\b\u0000G�IOɉ.�\fR\u0001\u0000v�\u0001\... \u0000\u0000\u0000\n�\u0004\u0000Philips Norelco Electric Shaver 8900, Wet & Dry Edition S8950-91.pngPK\u0001\u0002\u0014\u0000\u0014\u0000\u0000\u0000\b\u0000G�IO�\u0014\u000b��O\u0000\u0000L[\u0000\u0000\u0011\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000 \u0000\u0000\u0000x\u001d\u0006\u0000PR1342Prd1_HR.jpg

We are using `express-fileupload`  which gives us a `req.files` object and that includes the Buffer.

I believe also, the Content-Type is being set as `application/x-zip-compressed`

dpopp07:
@henhen87 Thanks for the issue. What exactly is the error you are getting? Also, will you provide a code sample so that we can understand how you are passing the Buffers?

It would also be helpful to know what version of the SDK you are using

henhen87:
Hello,

I am using `"ibm-watson": "^4.5.3",` and the Visual Recognition is `V3`. 

[...]

PR #985, opened 2019-10-14
fix(visual-recognition-v3): add dummy filename parameter

dpopp07:
Adds a dummy filename parameter to `createClassifier` positive examples so as to not break usage with the service when using Buffers. Since we can't read the filename from a Buffer, none gets sent. The service wants a filename, so behavior is messed up. The service doesn't actually ever use the filename, so a dummy here should be fine.

If this fixes behavior for #984, I will make this patch to v5 as well.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "ibm-watson",
-  "version": "4.5.1",
+  "version": "4.5.3",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


dpopp07:
@dardanos I'm not sure that this has been fixed before and I don't think that it relates to `axios`. It requires a handwritten change in the generated SDK code. If this fix solves the issue at hand, I will make it an official patch in version 5 and add a test.

germanattanasio:
@dpopp07 When Aamar was generating the SDK, and we were using `request`. He built a library to let users specify a file, filename, and content type and translate that into a `form-data` part. 

In some cases, when you use a stream, you want to, along with the filename, specify the `Content-Type`. Some of our services use the file extension, but some others use the `Content-Type`

PR #987, opened 2019-10-15
feat: improve typedoc documentation generation

SaltedCaramelCoffee:
##### Checklist

* Improved TypeDocs generation to hide interfaces, type aliases and modules from navigation and page
* Improved TypeDocs generation to expose first level properties of function/constructor arguments


ExiliumAeternum:
@dpopp07 The TSDoc specification shows [examples](https://github.com/Microsoft/tsdoc#what-is-tsdoc) to document only the parameters, not its first level properties. The extra parameters we put in the comment is ignored by TypeDoc since they are not used by the constructor.

Like you mentioned, since the type is from the core and not part of the SDK, I don't think TypeDoc found the definition of it and therefore it does not have its own page, and no link to the page either.

I've added logic to exclude sections that we are already excluding in the table of content. Please review the PR again.

PR #988, opened 2019-10-16
build: Add stale bot

germanattanasio:


.github/stale.yml
```diff
@@ -0,0 +1,17 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - security
```
dpopp07: We don't use either of these labels. Can we change this to be `bug`? I know bugs shouldn't last 60 days but I wouldn't want them to be closed just because no one has gotten to them. I'm fine to not do this, but its a suggestion.

```suggestion
  - bug
```

PR #1015, opened 2020-01-17
Restore model field in CategoriesOptions

mediumTaj:
This pull request restores a field that was inadvertently removed in the last regeneration cycle.


mediumTaj:
Yeah I don't think it's really breaking in Node... at least thats what I think Dustin would say, lol! I just wanted to get this in there for consistency as long as I'm not breaking anything!

dpopp07:
You know, that's what I thought too when I released the SDK because I noticed that in the diffs. Thinking about it now, it actually is _technically_ a breaking change so it's good that this is being added back in.

PR #1036, opened 2020-03-30
fix: update dependencies and examples

germanattanasio:
Updated a few dependencies and based on the CHANGELOGs. Dependencies are dropping Node 4 and 6 support. `async` moved to 3.0 but no changes for us.

I disable a breaking change in prettier and disable rules in the generated unit tests.


dpopp07:
Looks like integration tests are failing on a bad assert. Something must have changed in the service instance.

examples/.eslintrc.js
```diff
@@ -1,12 +1,12 @@
...
+  parserOptions: { ecmaVersion: 6 },
+  rules: {
+    'no-console': 'off',
+    'node/no-missing-require': 'off',
+    'require-jsdoc': 'off',
+    'valid-jsdoc': 'off',
+    'no-process-exit': 'off',
+    'prefer-const': 'off',
+    'no-var': 'off',
+  },
```
dpopp07, line 11: I guess our lint config should follow the lint rules, huh? 👍

examples/assistant.v1.js
```diff
@@ -6,25 +6,25 @@ var AssistantV1 = require('ibm-watson/assistant/v1');
  * Instantiate the Watson Assistant Service
  */
 var assistant = new AssistantV1({
-  username: process.env.ASSISTANT_USERNAME || '<assistant_username>',
-  password: process.env.ASSISTANT_PASSWORD || '<assistant_password>',
-  version: '2018-02-16'
+  // See: https://github.com/watson-developer-cloud/node-sdk#authentication
+  // iam_apikey: 'INSERT YOUR IAM API KEY HERE',
```
dpopp07, line 10: I think for the latest version, this should be `authenticator` and not `iam_apikey`

examples/assistant_tone_analyzer_integration/tone_assistant_integration.v1.js
```diff
@@ -66,8 +49,8 @@ var maintainToneHistoryInContext = true;
 var payload = {
   workspace_id: process.env.WORKSPACE_ID || '<workspace_id>',
```
dpopp07, line 50: The keys are now camel case: `workspaceId`

examples/assistant.v1.js
dardanos: does that mean that we need to add an authenticator object and then the apikey?

germanattanasio: does this mean we need to add an authentication field?

examples/browserify/.env.example
```diff
@@ -0,0 +1,2 @@
+TONE_ANALYZER_APIKEY=kcLLX123ghgjVSdM9Basdadasda90IesOrCIywWr0rNLo
```
dpopp07, line 1: Is this a fake key, right?

germanattanasio: > yeah don't you see the 123 or asdadasda9 😛

dpopp07: Didn't look that closely 😉

PR #1039, opened 2020-04-14
Release 2020 04 13

dpopp07:
Code regenerated to incorporate latest service updates.

Most notably, parameters were added to the HTTP and WebSocket `recognize` methods for Speech to Text.


visual-recognition/v4.ts
```diff
@@ -1914,7 +1914,7 @@ namespace VisualRecognitionV4 {
      */
     object: string;
     /** Number of bounding boxes in the collection with the updated object name. */
-    count: number;
+    count?: number;
```
mamoonraja, line 1917: any idea why this changed?

lib/recognize-stream.ts
```diff
@@ -177,6 +177,8 @@ class RecognizeStream extends Duplex {
       'redaction',
       'end_of_phrase_silence_time',
       'split_transcript_at_phrase_end',
+      'speech_detector_sensitivity',
```
mamoonraja, line 180: we need to add these here: https://github.com/watson-developer-cloud/node-sdk/blob/35fb2a5c9692c6e8a67f9de969269a8d37eb2ebb/lib/recognize-stream.ts#L105

visual-recognition/v4.ts
dpopp07: It changed in the API Definition

lib/recognize-stream.ts
dpopp07: Good catch, thanks

visual-recognition/v4.ts
mamoonraja, line 1917: Confirming if this will change after generating with new `API def`?

assistant/v2.ts
```diff
@@ -519,8 +519,20 @@ namespace AssistantV2 {
   export interface MessageContextSkill {
     /** Arbitrary variables that can be read and written by a particular skill. */
     user_defined?: JsonObject;
-    /** For internal use only. */
-    system?: JsonObject;
+    /** System context data used by the skill. */
+    system?: MessageContextSkillSystem;
```
mamoonraja, line 523: Does generating with the new API def tag still generate this change?

dpopp07: Re-generated the SDK, this change should not be present anymore

PR #1040, opened 2020-04-24
Update outdated links

mediumTaj:
### Summary
This pull request updates outdated links in documentation. Current redirects for these links will stop in May. This change has been made in the API definition.

### DIFF
--- a/assistant/v1.ts
+++ b/assistant/v1.ts
@@ -5336,7 +5336,7 @@ namespace AssistantV1 {
     message_to_human_agent?: string;
     /** The text of the search query. This can be either a natural-language query or a query that uses the Discovery
      *  query language syntax, depending on the value of the **query_type** property. For more information, see the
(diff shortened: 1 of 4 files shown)


assistant/v1.ts
```diff
@@ -5336,15 +5336,15 @@ namespace AssistantV1 {
     message_to_human_agent?: string;
     /** The text of the search query. This can be either a natural-language query or a query that uses the Discovery
      *  query language syntax, depending on the value of the **query_type** property. For more information, see the
-     *  [Discovery service documentation](https://cloud.ibm.com/docs/discovery/query-operators.html#query-operators).
+     *  [Discovery service documentation](https://cloud.ibm.com/docs/discovery?topic=discovery-query-operators#query-operators).
```
dpopp07, line 5339: Is this being addressed in the API Definition or does it need to be documented as a hand-written change?

mediumTaj: nope already addressed

PR #1052, opened 2020-08-05
fix: support `disableSslVerification` when user provides https agent

dpopp07:
Updates the core version to include a bug fix. This bug caused `disableSslVerification` to be ignored when users provided a custom https agent (which they would do if using a proxy, for example).

Code for the fix is here - https://github.com/IBM/node-sdk-core/pull/104

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1536,12 +1536,13 @@
       }
     },
     "@jest/types": {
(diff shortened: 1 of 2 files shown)


dpopp07:
This version of the core contains a bug. Will re-open once fixed

PR #1053, opened 2020-08-05
Release 2020 08 7

apaparazzi0329:
Regenerates the services, resulting in some models changing and four new methods: `listLogs` in Assistant v2, `analyzeDocument` and `updateCollection` in Discovery v2, and `listLanguages` in Language Translator v3. Integration tests are added for these new methods as well. Also, new voice options were added for Text to Speech v1

Lastly, every service had their default service url changed as well as several comment changes.

### DIFF
--- a/discovery/v2.ts
+++ b/discovery/v2.ts
@@ -20,10 +20,10 @@ import { Authenticator, BaseService, getAuthenticatorFromEnvironment, getMissing
 import { getSdkHeaders } from '../lib/common';
 
 /**
- * IBM Watson&trade; Discovery for IBM Cloud Pak for Data is a cognitive search and content analytics engine that you
- * can add to applications to identify patterns, trends and actionable insights to drive better decision-making.
- * Securely unify structured and unstructured data with pre-enriched content, and use a simplified query language to
- * eliminate the need for manual filtering of results.
+ * IBM Watson&trade; Discovery is a cognitive search and content analytics engine that you can add to applications to
+ * identify patterns, trends and actionable insights to drive better decision-making. Securely unify structured and
+ * unstructured data with pre-enriched content, and use a simplified query language to eliminate the need for manual
+ * filtering of results.
  */
 
 class DiscoveryV2 extends BaseService {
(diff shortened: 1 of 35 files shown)


test/integration/assistant.v2.test.js
```diff
@@ -116,4 +116,19 @@ describe('assistant v2 integration', () => {
       done();
     });
   });
+
+  it('should listLogs @slow', done => {
```
dpopp07, line 120: If this is one of the tests that requires a premium instance, you need to add a comment above describing that. Also, if I were to pull down this test and run it as is, with my credentials, my guess is that it would fail and I would be confused. So maybe we should comment it out entirely, or change the code to make it very clear that I would need new credentials to run it (like creating a specific instance of Assistant and using it only for this test, with the extra credentials needed).

visual-recognition/v3.ts
```diff
@@ -127,8 +127,8 @@ class VisualRecognitionV3 extends BaseService {
         },
         'url': _params.url,
         'threshold': _params.threshold,
-        'owners': Array.isArray(_params.owners) ? _params.owners.join(',') : _params.owners,
-        'classifier_ids': Array.isArray(_params.classifierIds) ? _params.classifierIds.join(',') : _params.classifierIds
```
dpopp07, line 131: These need to manually be changed back to normal after generating, I forgot to go over that with you

visual-recognition/v4.ts
```diff
@@ -118,8 +118,8 @@ class VisualRecognitionV4 extends BaseService {
       }
 
       const formData = {
-        'collection_ids': Array.isArray(_params.collectionIds) ? _params.collectionIds.join(',') : _params.collectionIds,
-        'features': Array.isArray(_params.features) ? _params.features.join(',') : _params.features,
+        'collection_ids': _params.collectionIds,
+        'features': _params.features,
```
dpopp07, line 122: These need to manually be changed back to normal after generating, I forgot to go over that with you

discovery/v2.ts
```diff
@@ -1177,15 +1554,852 @@ class DiscoveryV2 extends BaseService {
...
+        'customer_id': _params.customerId
+      };
+
+      const sdkHeaders = getSdkHeaders(DiscoveryV2.DEFAULT_SERVICE_NAME, 'v2', 'deleteUserData');
+
+      const parameters = {
+        options: {
+          url: '/v2/user_data',
+          method: 'DELETE',
+          qs: query,
```
dpopp07, line 2399: It looks like a ton of methods were added to this service but you only documented two of them in your commits. I don't think you need to change that at this point but in the future, you should document all of them

test/integration/discovery.v2.test.js
```diff
@@ -222,6 +241,30 @@ describe('discovery v2 integration @slow', () => {
...
+        res = await discovery.analyzeDocument(params);
+      } catch (err) {
+        return done(err);
+      }
+
+      expect(res).toBeDefined();
+      const { result } = res || {};
+      expect(result).toBeDefined();
+      done();
+    });
```
dpopp07, line 267: Looking at the discovery code, it looks like way more methods than just these two were added. Maybe I'm missing something but if not, we should add integration tests for all of the new methods as well

test/integration/assistant.v2.test.js
apaparazzi0329: Makes sense, I will make that change along with the api updates

test/integration/discovery.v2.test.js
apaparazzi0329: That is strange, I don't know how I missed all those new methods... I will make sure to write tests for them

mediumTaj:
Is this PR ready for release? Or are there still comments that need to be addressed?

dpopp07:
There are still comments that need to be addressed

test/integration/discovery.v2.test.js
```diff
@@ -38,7 +37,181 @@ describe('discovery v2 integration @slow', () => {
 
   let documentId;
 
+  const EXTENDED_TIMEOUT = 30000;
+
+  describe('projects', () => {
+    test(
+      'projects CRUD',
```
dpopp07, line 44: I see where you're coming from here, but I think we should hold to the testing principle that each test should cover functionality for one individual method. It is generally not good practice to include too much functionality in one test. If one of these requests were to fail, it would not be trivial to know which one - the failing test would just be `projects CRUD` and we would have to look at the line number in the code to see which operation failed.

Also, I should have mentioned this earlier but since you're writing new tests for a bunch of operations, we should clean up the format a little. The try/catch thing actually has some hidden potential for issues and it is not necessary. I updated the code in our Node template and you should probably match that format for the new tests (not required to backfit these changes to all tests at this point). [Here is a super bare-bones example](https://github.ibm.com/CloudEngineering/node-sdk-template/blob/master/test/integration/example-service.v1.test.js#L84-L90)

test/integration/discovery.v2.test.js
dpopp07: I think that a lot of the methods I saw that looked like they were added were old ones that just got moved (so they looked like additions). So I misinterpreted. You can check if there are already tests for some of these operations, just search in the code. No need to duplicate any testing

test/integration/discovery.v2.test.js
apaparazzi0329: I was honestly going based off Jeff's suggestion on how to make CRUD tests but I can certainly separate them 😅

dpopp07: Yeah I can see an argument for doing it that way, because I don't think tests are guaranteed to run in a certain order, but they always have run in the order we want them to so I think it's worth the risk to make the tests more granular.

PR #1058, opened 2020-09-24
fix: add `getTransactionId` method to the SynthesizeStream

dpopp07:
This method was present on the RecognizeStream and was mistakenly left off of the SynthesizeStream when it was developed. This PR fixes that mistake.

Rather than duplicate the method between the Stream classes, I moved the core functionality to a utility function that each Stream relies on. No code has actually changed; it works exactly the same.


apaparazzi0329:
I have already tested this and it works as expected. 👍🏼

apaparazzi0329 (approved):
Looks good

PR #1060, opened 2020-09-28
fix: stop ignoring serviceUrl for Websocket methods

### LINKED ISSUE #1059: recognizeUsingWebSocket does not support `serviceUrl` initialization parameter
**Overview**
The `recognizeUsingWebSocket` method from STT and TTS does not support the use of initializing `SpeechToTextV1` and `TextToSpeechV1` with `serviceUrl` but instead only supports `url`

**Expected behavior**
`recognizeUsingWebSocket` should support the use of `serviceUrl` as all other methods do

**Actual behavior**
A WebSocket connection error will be thrown when `serviceUrl` is used to initialize the service and the default url wss://stream.watsonplatform.net/speech-to-text/api  is not usable

**How to reproduce**
This error is only reproducible for watson instances that cannot use the default api url shown above
```
const STT_APIKEY = 'config.apikey';
const STT_URL = 'config.url';
const fs = require("fs");
const { IamAuthenticator } = require("ibm-watson/auth");
const SpeechToTextV1 = require("ibm-watson/speech-to-text/v1");


const speechToText = new SpeechToTextV1({
    authenticator: new IamAuthenticator({
        apikey: STT_APIKEY,
    }),
    serviceUrl: STT_URL,
    //url: STT_URL,
});

const params = {
    objectMode: true,
    contentType: 'audio/wav',
    model: 'it-IT_BroadbandModel',
    keywords: ['probabilmente', 'sono', 'visto'],
    keywordsThreshold: 0.5,
    maxAlternatives: 3,
    disableSslVerification: false,
};

// Create the stream.
const recognizeStream = speechToText.recognizeUsingWebSocket(params);

// Pipe in the audio.
fs.createReadStream('troy.wav').pipe(recognizeStream);

recognizeStream.pipe(fs.createWriteStream('transcription.txt'));

// Listen for events.
recognizeStream.on('data', function(event) { onEvent('Data:', event); });
recognizeStream.on('error', function(event) { onEvent('Error:', event); });
recognizeStream.on('close', function(event) { onEvent('Close:', event); });

// Display events on the console.
function onEvent(name, event) {
    console.log(name, JSON.stringify(event, null, 2));
};
```

**SDK Version**
5.7.0

dpopp07:
The Websocket methods `recognizeUsingWebsocket` (STT) and `synthesizeUsingWebsocket` (TTS) were not using the `serviceUrl` parameter defined by the core. They were only looking for `url`.

The `serviceUrl` parameter is the correct one, `url` is deprecated (but still supported). In the core, `serviceUrl` and `url` are both mapped to `baseOptions.serviceUrl`, so we only need to look at that single parameter and both are supported. This is a compatible change. I updated the test to make sure the user can provide both `serviceUrl` and `url`.

Resolves #1059

### DIFF
--- a/lib/recognize-stream.ts
+++ b/lib/recognize-stream.ts
@@ -69,7 +69,7 @@ class RecognizeStream extends Duplex {
    *
    * @param {Options} options
    * @param {Authenticator} options.authenticator - Authenticator to add Authorization header
-   * @param {string} [options.url] - Base url for service (default='wss://stream.watsonplatform.net/speech-to-text/api')
+   * @param {string} [options.serviceUrl] - Base url for service (default='wss://stream.watsonplatform.net/speech-to-text/api')
    * @param {OutgoingHttpHeaders} [options.headers] - Only works in Node.js, not in browsers. Allows for custom headers to be set, including an Authorization header (preventing the need for auth tokens)
    * @param {boolean} [options.readableObjectMode] - Emit `result` objects instead of string Buffers for the `data` events. Does not affect input (which must be binary)
    * @param {boolean} [options.objectMode] - Alias for readableObjectMode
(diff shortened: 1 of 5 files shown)


apaparazzi0329 (approved):
These changes fix the issue stated and are good to go 👍🏼

PR #1066, opened 2020-10-12
docs: add special note regarding use of credentials file with assistant

apaparazzi0329:
This PR adds a special note for using the credentials file authentication option with assistant. Legacy issues prevent this from being a quick fix but the README should at least address this issue.

##### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,10 +162,7 @@ const assistant = new VisualRecognitionV3({
   serviceName: 'visual-recognition',
 })
 ```
-It is worth noting that implementing this will prioritize reading the credentials file over using the `VCAP_SERVICES` variable when pushing to IBM Cloud. This will become an issue if you do not push a credentials file to the cloud as VCAP does not know how to interpret these `serviceName` parameters. Credential priority follows this order: 
-1. Programmatic (i.e. IamAuthenticator)
-2. Credentials File
-3. VCAP
+It is worth noting that if you are planning to rely on VCAP_SERVICES for authentication then the `serviceName` parameter **MUST** be removed otherwise VCAP_SERVICES will not be able to authenticate you. See [Cloud Authentication Prioritization](#cloud-authentication-prioritization) for more details.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -147,6 +147,15 @@ And that's it!
 
 If you're using more than one service at a time in your code and get two different `ibm-credentials.env` files, just put the contents together in one `ibm-credentials.env` file and the SDK will handle assigning credentials to their appropriate services.
 
+**Special Note**: Due to legacy issues in Assistant V1 and V2 the following parameter `serviceName` must be added when creating the service object:
+```js
+const AssistantV1 = require('ibm-watson/assistant/v1');
+const assistant = new AssistantV1({
+  version: '2020-04-01',
+  serviceName: 'assistant'
```
dpopp07, line 155:
```suggestion
  serviceName: 'assistant',
```

dpopp07 (changes requested):
I completely forgot to mention to you that this is also a problem for Visual Recognition - will you add another snippet below for that?

Also, I think it's worth noting that this will impact how the `VCAP_SERVICES` variable is read on IBM Cloud so users aren't surprised if they leave this in and it causes issues on the cloud

README.md
```diff
@@ -147,6 +147,26 @@ And that's it!
...
+})
+```
+```js
+const VisualRecognitionV3 = require('ibm-watson/visual-recognition/v3');
+const assistant = new VisualRecognitionV3({
+  version: '2018-03-19',
+  serviceName: 'visual-recognition',
+})
+```
+It is worth noting that implementing this will prioritize reading the credentials file over using the `VCAP_SERVICES` variable when pushing to IBM Cloud. This will become an issue if you do not push a credentials file to the cloud as VCAP does not know how to interpret these `serviceName` parameters. Credential priority follows this order:
```
dpopp07, line 165: I think there was a slight miscommunication here. The `serviceName` parameter does not impact priority at all so it won't specifically prioritize the credentials file. The VCAP reader works fine with the service name. The problem is, the VCAP object will never have keys for "assistant" or "visual-recognition". So, if the user intends to use the VCAP object on the cloud, they should not specify the service name

dpopp07 (approved):
Looks good 👍

PR #1072, opened 2020-11-16
rc 6.0.0

apaparazzi0329:
This PR is the release candidate branch for the upcoming major release version 6.0.0


MIGRATION-V6.md
```diff
@@ -0,0 +1,146 @@
...
+
+## Breaking changes
+
+#### Promises
+The detailed response is always returned. The `return_response` parameter is removed.
+```js
+const response = await listWorkspaces();
+console.log(response.result); // prints the body
+console.log(response.headers); // prints the headers
+```
```
dpopp07, line 20: Didn't this happen in the last major release? We don't need to document it here if so

MIGRATION-V6.md
```diff
@@ -0,0 +1,146 @@
...
+```js
+//Promise
+const assistant = new AssistantV2(options);
+try{
+  await res = assistant.message(params);
+  console.log(res.result) //Print out body
+} catch(err){
+  console.log(err) //Print out error
+}
```
dpopp07, line 43: You might want to show use of the `then` pattern here, just for its similarity to the callback pattern. Then, you can also show the `async/await` pattern. For that case, I would remind the user that `await` must be used inside of an `async` function or just show the function definition in your example.

MIGRATION-V6.md
```diff
@@ -0,0 +1,146 @@
...
+
+```
+
+### Support for Node v10 Dropped
+The SDK no longer supports Node version 10, as reflected in the `engines` property in the package.json file. Version 6 reached end of life in April 2019 and Version 8 reaches end of life on 31 December 2019.
```
dpopp07, line 48: We're only dropping v10 so there's no need to document when v6 and v8 ended. Replace that with the end date of v10

MIGRATION-V6.md
```diff
@@ -0,0 +1,146 @@
...
+##### Speech to Text
+- `recognize()` and `createJob()` now support Canadian French broadband and narrowband models
+- `createLanguageModel()` and `createAcousticModel()` now supports Australian English broadband and narrowband models
+- `addGrammar()` parameter `grammarFile` now supports `ReadableStream` and `Buffer`
+
+#### Visual Recognition v4
+- `createCollection()` parameter `trainingStatus` added: Training status information for the collection.
+- `updateCollection()` parameter `trainingStatus` added: Training status information for the collection.
+- Interface `CollectionTrainingStatus` added: Training status information for the collection.
+- Interface `ObjectDetailLocation` added: Defines the location of the bounding box around the object.
```
dpopp07, line 146: I'm not going to look at all of these in detail. The format looks good, just make sure that you aren't including anything from the previous release on accident

assistant/v1.ts
```diff
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright IBM Corp. 2018, 2020.
+ * (C) Copyright IBM Corp. 2020.
```
dpopp07, line 2: Make sure you go through all of the copyrights before merging

visual-recognition/v3.ts
```diff
@@ -111,59 +119,46 @@ class VisualRecognitionV3 extends BaseService {
...
-      const formData = {
-        'images_file': {
-          data: _params.imagesFile,
-          filename: _params.imagesFilename,
-          contentType: _params.imagesFileContentType
-        },
-        'url': _params.url,
-        'threshold': _params.threshold,
-        'owners': Array.isArray(_params.owners) ? _params.owners.join(',') : _params.owners,
-        'classifier_ids': Array.isArray(_params.classifierIds) ? _params.classifierIds.join(',') : _params.classifierIds
```
dpopp07, line 131: @mediumTaj has this finally been fixed in the API definition or do we still need manual support for Array types here?

visual-recognition/v4.ts
```diff
@@ -99,64 +107,47 @@ class VisualRecognitionV4 extends BaseService {
...
-        if (_callback) {
-          _callback(missingParams);
-          return resolve();
-        }
-        return reject(missingParams);
-      }
-
-      const formData = {
-        'collection_ids': Array.isArray(_params.collectionIds) ? _params.collectionIds.join(',') : _params.collectionIds,
-        'features': Array.isArray(_params.features) ? _params.features.join(',') : _params.features,
```
dpopp07, line 122: See my comment on the v3 SDK

speech-to-text/v1.ts
```diff
@@ -49,77 +50,61 @@ class SpeechToTextV1 extends GeneratedSpeechToTextV1 {
...
-        const result = res;
-        if (err) {
-          return callback(err);
+      }
+    );
+
+    return async.retry(
+      options,
+      async (done) => {
+        try{
```
dpopp07, line 89: This all looks good, there are just some style things. For example, there should be a space before the bracket on the try and the catch. A lot of statements are also missing semicolons.

```suggestion
        try {
```

speech-to-text/v1.ts
```diff
@@ -49,77 +50,61 @@ class SpeechToTextV1 extends GeneratedSpeechToTextV1 {
...
+          const res = await self.listCorpora(params)
```
dpopp07, line 90:
```suggestion
          const res = await self.listCorpora(params);
```

I'm only leaving a comment here as a reminder but a lot of statements need semicolons at the end to conform to our style guide

dpopp07 (changes requested):
This is looking good. There are a few things to fix before merging, re-tag me once these are done and I'll give it another look over

visual-recognition/v3.ts
dpopp07: Actually Mike just said that he updated the code in the service to accept either an array or a comma-separated string. In that case, we don't need this change anymore. Is that true, Taj?

assistant/v1.ts
apaparazzi0329: Good catch

dpopp07 (commented):
Thanks for fixing that stuff up. I've left a number of comments on the tests that are not ultimately _required_. My thinking is that if you are going through the process of modernizing all of the tests, it might be worth cleaning up a lot of the sloppy testing practices that exist. I've left a bunch of comments on the Assistant v1 test showing what I mean. I'll leave it up to you to decide how much you want to address

visual-recognition/v3.ts
mediumTaj: Nope, we still need this as comma separated strings. Mike updated in visrec v4 but they didn't think it was worth it to update v3.

visual-recognition/v4.ts
mediumTaj: funny thing about v4 - it supports these as both comma separated strings and separate form parts. I think they are generated as separate form parts so no handwritten change needed anymore for this.

MIGRATION-V6.md
```diff
@@ -0,0 +1,146 @@
...
+- Interface `FeaturesResultsMetadata` added: Webpage metadata, such as the author and the title of the page.
+
+##### Speech to Text
+- `recognize()` and `createJob()` now support Canadian French broadband and narrowband models
+- `createLanguageModel()` and `createAcousticModel()` now supports Australian English broadband and narrowband models
+- `addGrammar()` parameter `grammarFile` now supports `ReadableStream` and `Buffer`
+
+#### Visual Recognition v4
+- `createCollection()` parameter `trainingStatus` added: Training status information for the collection.
+- `updateCollection()` parameter `trainingStatus` added: Training status information for the collection.
```
mediumTaj, line 152: training  status should be removed with a handwritten change. This should be readonly

personality-insights/v3.ts
```diff
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright IBM Corp. 2017, 2020.
+ * (C) Copyright IBM Corp. 2020.
```
mediumTaj: Did we run both lint and fix copyrights here? there should be two dates

visual-recognition/v3.ts
dpopp07: Thanks for the updates Taj! So Angelo, you'll need to retain this code for the major release.

personality-insights/v3.ts
apaparazzi0329: I did run both but it appears the copyright linter missed this one. I'll fix it manually and make sure to check every file

MIGRATION-V6.md
apaparazzi0329: Thank you for catching that

speech-to-text/v1.ts
```diff
@@ -49,77 +49,61 @@ class SpeechToTextV1 extends GeneratedSpeechToTextV1 {
...
+      const result = res.result;
+      if (!result.corpora.length) {
+        const sttError: SpeechToTextV1.SpeechToTextError = new Error(
+          'Customization has no corpa and therefore corpus cannot be analyzed'
+        );
+        sttError.code = SpeechToTextV1.ERR_NO_CORPORA;
+        return Promise.reject<any>(sttError);
+      }
+    } catch(err) {
+      return Promise.reject<any>(err)
```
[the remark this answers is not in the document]
apaparazzi0329: Nothing gets past your eagle eyes 😂

.travis.yml
```diff
@@ -6,17 +6,18 @@ node_js:
...
   npm: false
 before_install:
-- '[ "${TRAVIS_PULL_REQUEST}" = "false" ] && [ ! -z ${encrypted_ac3aacad7ba8_key} ] &&
-  openssl aes-256-cbc -K $encrypted_ac3aacad7ba8_key
-  -iv $encrypted_ac3aacad7ba8_iv -in secrets.tar.enc -out test/resources/secrets.tar
-  -d || true'
-- '[ "${TRAVIS_PULL_REQUEST}" = "false" ] && [ ! -z ${encrypted_ac3aacad7ba8_key} ]
-  && cd test/resources/ && tar xvf secrets.tar
-  && cd ../.. || true'
+- openssl aes-256-cbc -K $encrypted_02d9f3c197ce_key -iv $encrypted_02d9f3c197ce_iv
```
dpopp07, line 9: You may want to confirm this because I could be wrong - but by removing the "if" statement about it not being a pull request, it looks like the `auth.js` file will be created and that would mean the integration tests would be run. We do not want to run the integration tests on PR builds, only Push builds.

Whenever I updated the credentials, I would just drop the changes to the `.travis.yml` file - they are just formatting changes and are not needed. You can probably do the same here. The only thing I'm not sure about is this new key (`$encrypted_02d9f3c197ce_key`) - I'm not sure why that's there or how you need to included it

dpopp07 (approved):
Looks good! 👍 

As I mentioned before, there are some issues with the async/await pattern in the tests. These are not urgent and they can be addressed post-release. I'll help fix those up soon

PR #1091, opened 2021-04-01
fix: update core to enable use with more front-end environments

dpopp07:
This includes a version of the core package that stubs out modules that
are both unnecessary for and problematic in certain browser environments
like React.

See this commit for more info:
https://github.com/IBM/node-sdk-core/commit/006991bfe89c0671f9c053457cebe52cec924fed


apaparazzi0329:
LGTM

apaparazzi0329:
Travis credentials will need to be updated for STT to pass CI

apaparazzi0329 (approved):

PR #1094, opened 2021-04-22
ci: 9260 gha

Mikemosca:
Changes to add github actions


.github/workflows/build-test.yml
```diff
@@ -0,0 +1,54 @@
...
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+  build_test:
+    name: Build and Test on Node ${{ matrix.node-version }} and ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        node-version: ['10.x', '12.x']
```
repjarms, line 24: we should probably include version `14.x` since that is the current LTS from Node.

repjarms (approved):
LGTM, although I made one comment about expanding the node version matrix to include node 14.

.github/workflows/build-test.yml
Mikemosca: yea, sounds good.  how about I put that in with the next PR to turn on semantic release?

PR #1096, opened 2021-05-18
Release 2021 05 14

apaparazzi0329:


speech-to-text/v1.ts
```diff
@@ -279,6 +279,7 @@ namespace SpeechToTextV1 {
     splitTranscriptAtPhraseEnd?: boolean;
     speechDetectorSensitivity?: number;
     backgroundAudioSuppression?: number;
+    lowLatency?: boolean;
```
dpopp07, line 282: This needs to be accompanied by changes in `lib/recognize-stream.ts` in order for the new parameter to actually be supported

assistant/v1.ts
```diff
@@ -5198,9 +5336,11 @@ namespace AssistantV1 {
     agent_unavailable?: AgentAvailabilityMessage;
     /** Routing or other contextual information to be used by target service desk systems. */
     transfer_info?: DialogNodeOutputConnectToAgentTransferInfo;
+    /** An array of objects specifying channels for which the response is intended. */
+    channels?: ResponseGenericChannel[];
   }
 
-  /** An object that describes a response with response type `image`. */
+  /** DialogNodeOutputGenericDialogNodeOutputResponseTypeImage. */
```
dpopp07, line 5343: It's interesting that we're seeing these changes from an actual description to what looks like an ID. It doesn't seem very desirable to me. This is coming from the generator, right? I wonder if the descriptions were removed from the API definition.

It only seems to be happening in Assistant (v1 and v2)

test/integration/assistant.v1.test.js
```diff
@@ -1066,4 +1066,20 @@ describe('assistant v1 integration', () => {
...
+        input: {
+          text: 'Turn on the lights',
+        },
+        workspaceId: options.workspaceId,
+      };
+
+      const res = await assistant.bulkClassify(params);
+      const { result } = res || {};
+      expect(result).toBeDefined();
+      console.log(result);
```
dpopp07, line 1082: Is this console log intentional? Or leftover from testing?

dpopp07 (changes requested):
Looks good for the most part. The hand-written websocket code needs to be updated and and I left a couple other comments about things that looked unusual

speech-to-text/v1.ts
apaparazzi0329: Thank you, I did not realize this

test/integration/assistant.v1.test.js
apaparazzi0329: Good catch that will be removed

assistant/v1.ts
apaparazzi0329: Yes looking at the previous version of the api refs it looks like the description for these discriminator models was removed. As to why, I would not know but it had to be done on purpose

dpopp07: Yeah that's weird. Nothing that should hold up the release though

PR #1099, opened 2021-06-02
fix(tts): remove unnecessary filename param

apaparazzi0329:


dpopp07 (approved):
Looks good to me. I'm assuming this will be reflected in the API definition at some point, since this is generated code?

apaparazzi0329:
It is already reflected in the API defs

PR #1111, opened 2021-09-08
Hsaylor 9718 rc

Hsaylor:
Generated SDK release using :
API Defs tag `sdk-2021-09-03`
Generator tag `3.38.0`

Discovery V1 tag `sdk-2021-09-03`

# Service Changes
### Assistant V1
* `alt_text` property added to Image response type
​
### Assistant V2
* `session_start_time` and `state`  properties added to `MessageContextGlobalSystem` model
* `alt_text` property added to Image response type
* `location` property of `RuntimeEntity` model changed from required to optional
​
### Discovery V2
* Enum update for  `CreateProjectConstants`
​
### NLU
* Fix for `listClassificationsModels` through return type change and removal of unnecessary model
​
### STT
* More languages supported for next generation models
​
### TTS
* New voice models added
​
### CC, Disco V1, LT, NLC, PI, TA, VisRec V3 & V4
* No changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,10 +55,11 @@
     "@commitlint/config-conventional": "^8.2.0",
     "@semantic-release/changelog": "^5.0.1",
     "@semantic-release/git": "^9.0.0",
-    "axios": "^0.21.1",
+    "axios": "^0.21.4",
     "browserify": "^16.5.2",
     "codecov": "^3.8.1",
     "concat-stream": "^2.0.0",
+    "cz-conventional-changelog": "^3.3.0",
     "eslint": "^6.8.0",
     "eslint-config-google": "^0.14.0",
     "eslint-config-prettier": "^6.15.0",
@@ -202,5 +203,10 @@
       ]
     }
   },
-  "snyk": true
+  "snyk": true,
+  "config": {
+    "commitizen": {
+      "path": "./node_modules/cz-conventional-changelog"
+    }
+  }
 }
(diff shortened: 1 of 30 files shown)


package.json
```diff
@@ -55,10 +55,11 @@
     "@commitlint/config-conventional": "^8.2.0",
     "@semantic-release/changelog": "^5.0.1",
     "@semantic-release/git": "^9.0.0",
-    "axios": "^0.21.1",
+    "axios": "^0.21.4",
     "browserify": "^16.5.2",
     "codecov": "^3.8.1",
     "concat-stream": "^2.0.0",
+    "cz-conventional-changelog": "^3.3.0",
```
apaparazzi0329, line 62: Why was this dependency added?

apaparazzi0329: Also, I believe there was an ask to upgrade the ibm-cloud-sdk-core dependency to the latest version 2.14.3. While the current notation would still install the latest version we don't want users the ability to install a version of the ibm-cloud-sdk-core that still uses a vulnerable version of axios so we should be specific

apaparazzi0329 (changes requested):
Looks good. Requested a small change regarding our dependencies

package.json
Hsaylor: `cz-conventional-changelog` was added as a fix to the commitizen pipeline. We recommend it's use but weren't including it. It made sense to add it while we were updating the `package.json`. I got confirmation from @repjarms that it should be included. 

Ah good catch, I updated our `axios` dependency past the vulnerable version, but missed the update on the cloud core itself

apaparazzi0329 (approved):
LGTM!

apaparazzi0329 (approved):
Changes LGTM

PR #1112, opened 2021-09-17
fix(stt): timestamps type to properly reflect response model

paulchiu:
Commit https://github.com/watson-developer-cloud/node-sdk/commit/84bb18da2a4431c4f59cd95f0e05c8d61cd7ab85 rolled back PR https://github.com/watson-developer-cloud/node-sdk/pull/1085, I believe this was unintentional.

I have verified that the response type is still `[string, number, number]`, which reflects the current TSDoc too.

This change re-applies the fix created by @apaparazzi0329.


apaparazzi0329:
Yes you are correct this was an unintentional change. Thank you for catching this and opening up a PR. We will look to merge your PR and release a patch for this on Monday

Hsaylor (approved):

PR #1113, opened 2021-09-21
fix(stt): revert change to word_confidence made by 84bb18d

Hsaylor:
Second PR to fix STT `word_confidence` as a follow up to #1112

### DIFF
--- a/speech-to-text/v1-generated.ts
+++ b/speech-to-text/v1-generated.ts
@@ -6153,7 +6153,7 @@ namespace SpeechToTextV1 {
      *  `[["hello",0.95],["world",0.866]]`. Confidence scores are returned only for the best alternative and only with
      *  results marked as final.
      */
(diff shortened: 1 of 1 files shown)


repjarms:
@Hsaylor did we decide if we were going to do this? If I remember correctly, this is something we aren't going to override at this time. If that's the case we should close this.

apaparazzi0329:
Well I looked over the `timestamp` and `word_confidence` properties and they are both currently incorrect, including this PR. We can create a quick patch if we think this needs to be addressed but the correct typing for each property, respectively, is `[string, number, number][]` and `[string, number][]`

apaparazzi0329:
@Hsaylor let's do a quick patch release to address this

Hsaylor:
Update pushed

apaparazzi0329 (approved):
lgtm

PR #1114, opened 2021-09-23
fix(stt): timestamps type to properly reflect response model

paulchiu:
My apologies, I made a typo in commit https://github.com/watson-developer-cloud/node-sdk/pull/1112/commits/1efa98382303fda77ba078541e0c7922d750339d.

When copying and pasting the new type I overwrote the `[]` at the end. This means that the type is still incorrect.


rorybe (approved):

apaparazzi0329:
Thank you for bringing this to our attention. We also noticed that the word_confidence property was incorrectly typed. These have both been addressed in this commit: 9f4d6645fa8080b41f4f99fb0cf07e3f5b68a4c1

PR #1117, opened 2021-11-04
ci: 9879 inttest

Mikemosca:
Add a GHA integration test workflow


.github/workflows/integration-test.yml
```diff
@@ -0,0 +1,111 @@
...
+        VISUAL_RECOGNITION_APIKEY: ${{ secrets.VR_APIKEY }}
+        VISUAL_RECOGNITION_COLLECTION_ID: ${{ secrets.VR_COLLECTION_ID }}
+        DISCOVERY_URL: "https://api.us-south.discovery.watson.cloud.ibm.com"
+        DISCOVERY_APIKEY: ${{ secrets.D1_APIKEY }}
+        DISCOVERY_ENVIRONMENT_ID: ${{ secrets.D1_ENVIRONMENT_ID }}
+        DISCOVERY_CONFIGURATION_ID: ${{ secrets.D1_CONFIGURATION_ID }}
+        DISCOVERY_COLLECTION_ID1: ${{ secrets.D1_COLLECTION_ID1 }}
+        DISCOVERY_COLLECTION_ID2: ${{ secrets.D1_COLLECTION_ID2 }}
+        DISCOVERY_COLLECTION_ID3: ${{ secrets.D1_COLLECTION_ID3 }}  # japanese
+
```
apaparazzi0329, line 74: LGTM, my only question is if Discovery V2 uses its own set of credentials because I was under the impression that it did

apaparazzi0329 (approved):

.github/workflows/integration-test.yml
Mikemosca: Your right,  I need to fix this up.  I will do that now and let you know if another review is needed.  Thank you

PR #1177, opened 2022-10-24
fix(deps): upgrade core version to address file-type vulnerability

apaparazzi0329:
stt: fix "waiting" methods so they no longer send invalid params

### DIFF
--- a/test/integration/speech-to-text.test.js
+++ b/test/integration/speech-to-text.test.js
@@ -154,7 +154,7 @@ describe('speech to text integration', () => {
         customizationId,
         corpusName: 'test_corpus_1',
         corpusFile: fs.createReadStream(
-          path.join(__dirname, '../resources/speech_to_test/corpus-short-1.txt')
+          path.join(__dirname, '../resources/speech_to_text/corpus-short-1.txt')
         ),
       };
       await speechToText.addCorpus(params);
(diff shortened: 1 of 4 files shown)


test/integration/speech-to-text.test.js
```diff
@@ -177,9 +177,9 @@ describe('speech to text integration', () => {
         customizationId,
         corpusName: 'test_corpus_3',
         corpusFile: fs
-          .readFileSync(path.join(__dirname, '../resources/speech_to_test/corpus-short-2.txt'))
+          .readFileSync(path.join(__dirname, '../resources/speech_to_text/corpus-short-2.txt'))
           .toString(),
-        allow_overwrite: true,
+        allowOverwrite: true,
```
jeff-arn, line 182: this feels like a breaking change

apaparazzi0329: This is fixing a test that was failing because the actual param is `allowOverwrite`. This was not picked up before because the node-sdk did not validate params in the past

jeff-arn: Ok sounds good

jeff-arn (approved):
LGTM - merge when ready

Hsaylor (approved):
LGTM

PR #1186, opened 2023-01-12
fix(deps): upgrade core version to address jsonwebtoken vulnerability

apaparazzi0329:


test/integration/discovery.v1.test.js
```diff
@@ -93,7 +93,6 @@ describe('discovery v1 integration', () => {
   it('should listCollections()', async () => {
     const params = {
       environmentId,
-      configurationId: configurationId,
```
jeff-arn, line 96: why is `configurationId` removed?

test/integration/discovery.v1.test.js
```diff
@@ -105,12 +104,10 @@ describe('discovery v1 integration', () => {
   it('should perform a federated query for notices', async () => {
     const params = {
       environmentId,
-      configurationId: configurationId,
```
jeff-arn, line 108: same as above

test/integration/discovery.v1.test.js
```diff
@@ -105,12 +104,10 @@ describe('discovery v1 integration', () => {
...
       collectionIds: [collectionId, collectionId2],
       filter: 'yesplease',
       count: 10,
       sort: ['+field_1', '-field_2'],
-      natural_language_query: 'a question about stuff and things',
```
jeff-arn, line 113: why is `natural_language_query` removed?

test/integration/discovery.v1.test.js
```diff
@@ -405,7 +402,6 @@ describe('discovery v1 integration', () => {
       expect(Array.isArray(result.aggregations)).toBe(true);
       expect(result.aggregations[0].results).toBeDefined();
       expect(Array.isArray(result.aggregations[0].results)).toBe(true);
-      expect(result.aggregations[0].results[0].matching_results).toBeDefined();
```
jeff-arn, line 408: why is `matching_results` removed?

test/integration/discovery.v1.test.js
```diff
@@ -418,7 +414,6 @@ describe('discovery v1 integration', () => {
       expect(Array.isArray(result.aggregations)).toBe(true);
       expect(result.aggregations[0].results).toBeDefined();
       expect(Array.isArray(result.aggregations[0].results)).toBe(true);
-      expect(result.aggregations[0].results[0].event_rate).toBeDefined();
```
jeff-arn, line 421: why is `event_rate` removed?

test/integration/text-to-speech.test.js
```diff
@@ -53,7 +54,14 @@ describe('text to speech_integration', () => {
       });
 
       synthStream.on('close', (code, reason) => {
-        done();
+        console.log(code);
```
jeff-arn, line 57: remove `console.log

test/integration/text-to-speech.test.js
```diff
@@ -53,7 +54,14 @@ describe('text to speech_integration', () => {
...
+        if (code === 1000) {
+          done();
+        } else if (code === undefined) {
+          // Do nothing: this is in order to prevent done() being called twice
```
jeff-arn, line 61: I see the comment inline, but can you elaborate more on this? It seems that normally if `code === undefined` we would go to the `else` block and call `fail();`. Do we not want to fail the test if it is `undefined`?

jeff-arn (commented):
I left some questions inline on the PR

jeff-arn:
@apaparazzi0329 also can you update the Readme in this PR to correct what the minimum version of node is?

test/integration/discovery.v1.test.js
apaparazzi0329: This is no longer a valid parameter

test/integration/discovery.v1.test.js
apaparazzi0329: This is no longer a valid parameter

test/integration/text-to-speech.test.js
apaparazzi0329: Whoops good catch

test/integration/text-to-speech.test.js
apaparazzi0329: Normally yes but this undefined is actually a duplicate call which creates a duplicate `done` call which is no longer valid

test/integration/discovery.v1.test.js
apaparazzi0329: event_rate is an optional property and is null in this case

test/integration/text-to-speech.test.js
jeff-arn: Why is there a duplicate call on the `close` event?

apaparazzi0329: Your guess is as good as mine but I did test the current version and this has already existed and only came up now due to updating Jest

jeff-arn: Hmm ok

jeff-arn (approved):
LGTM

PR #1187, opened 2023-03-01
SDK major release 2023

apaparazzi0329:
Note: READMEs still need to be updated


discovery/v2.ts
```diff
@@ -5412,6 +5399,15 @@ namespace DiscoveryV2 {
     notices?: Notice[];
   }
 
+  /** Result for the `pair` aggregation. */
+  export interface QueryPairAggregationResult {
```
kevinkowa, line 5403: weird, I don't see this being used anywhere

kevinkowa (approved):
Looks good, just one comment

discovery/v2.ts
apaparazzi0329: It is used in this interface: `QueryAggregationQueryPairAggregation`

PR #1203, opened 2024-02-21
Sdk minor release 2024

apaparazzi0329:


assistant/v2.ts
```diff
@@ -196,11 +190,7 @@ class AssistantV2 extends BaseService {
       'include_audit': _params.includeAudit,
     };
 
-    const sdkHeaders = getSdkHeaders(
-      AssistantV2.DEFAULT_SERVICE_NAME,
-      'v2',
-      'listAssistants'
-    );
+    const sdkHeaders = getSdkHeaders(AssistantV2.DEFAULT_SERVICE_NAME, 'v2', 'listAssistants');
```
jeff-arn, line 193: are these indent changes a deliberate stylistic change or are we forgetting to run something like prettier?

jeff-arn (commented):
Left a comment about style

assistant/v2.ts
apaparazzi0329: They are a deliberate stylistic change from the sdk-generator

jeff-arn (approved):

